PR #35, opened 2014-07-11
AddOn deletes fail when done from a list of plan add-ons

hross:
The AddOn list that gets loaded from plan.AddOns returns AddOn objects that don't have a planCode associated with them, which causes certain operations to fail (Delete's in particular). I updated the AddOnList to store its associated plan code and update retrieved add-ins with this data.

### DIFF
--- a/Library/AddOn.cs
+++ b/Library/AddOn.cs
@@ -35,8 +35,9 @@ namespace Recurly
         {
         }
 
-        internal AddOn(XmlTextReader xmlReader)
+        internal AddOn(string planCode, XmlTextReader xmlReader)
         {
+            PlanCode = planCode;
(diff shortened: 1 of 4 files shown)


bhelx:
I believe these are related:

https://github.com/recurly/recurly-client-net/pull/141

bhelx:
After reviewing, https://github.com/recurly/recurly-client-net/pull/141 only takes care of the GetAddOn whereas this takes care of that method and the AddOnList side.

I'm going to keep this open until I implement https://github.com/recurly/recurly-client-net/issues/142 which will take care of all of these problems library wide.

aaron-suarez:
We're not going to implement the HREFs at this time for this library. Closing this and the related issue.

PR #64, opened 2015-01-05
Adding vat_location_valid field to Account

bhelx:
This field will show up when the account is elligible and has been
location verified according to the VAT rules. It will be true when
passed and false when failed.


Library/Account.cs
```diff
@@ -38,6 +38,7 @@ public enum AccountState : short
         public string AcceptLanguage { get; set; }
         public string HostedLoginToken { get; private set; }
         public DateTime CreatedAt { get; private set; }
+        public bool VatLocationValid { get; private set; }
```
bhelx: Is there a way to not generate a setter?

PR #67, opened 2015-01-15
Adding notes support for Invoice and Subscription

bhelx:
Trying to catch .NET up to the other libraries with notes support.

#### Testing

You can run this and verify the notes have changed. If you run again you should be able to see they have persisted.

```csharp
var sub = Subscriptions.Get("2bf76b61f7874bb58c384d45759be62c");

Console.WriteLine(sub.CustomerNotes);
Console.WriteLine(sub.TermsAndConditions);
Console.WriteLine(sub.VatReverseChargeNotes);

Dictionary<string, string> notes = new Dictionary<string, string>();
notes.Add("CustomerNotes", "New Customer Notes");
notes.Add("TermsAndConditions", "New T and C");
notes.Add("VatReverseChargeNotes", "New VAT Notes");
sub.UpdateNotes(notes);

Console.WriteLine(sub.CustomerNotes);
Console.WriteLine(sub.TermsAndConditions);
Console.WriteLine(sub.VatReverseChargeNotes);
```

### DIFF
--- a/Library/Subscription.cs
+++ b/Library/Subscription.cs
@@ -200,6 +200,10 @@ namespace Recurly
         /// </summary>
         internal bool _saved;
 
+        public string CustomerNotes { get; set; }
+        public string TermsAndConditions { get; set; }
+        public string VatReverseChargeNotes { get; set; }
+
         internal Subscription()
         {
             IsPendingSubscription = false;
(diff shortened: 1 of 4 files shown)


Library/Subscription.cs
```diff
@@ -341,6 +345,29 @@ public void Postpone(DateTime nextRenewalDate)
                 ReadXml);
         }
 
+        public bool UpdateNotes(String customerNotes, String termsAndConditions, String vatReverseChargeNotes)
```
bhelx: Should we be passing in some kind of hasmap here?

Library/Subscription.cs
```diff
@@ -341,6 +345,29 @@ public void Postpone(DateTime nextRenewalDate)
...
+        {
+
+            if (customerNotes != null)
+                CustomerNotes = customerNotes;
```
bhelx: Not sure if it's the best idea to set this on the client side subscription object. What would/should happen if the API call fails?

Library/Subscription.cs
```diff
@@ -341,6 +345,29 @@ public void Postpone(DateTime nextRenewalDate)
...
+            if (customerNotes != null)
+                VatReverseChargeNotes = vatReverseChargeNotes;
+
+            Client.Instance.PerformRequest(Client.HttpRequestMethod.Put,
+                UrlPrefix + Uri.EscapeUriString(Uuid) + "/notes",
+                WriteSubscriptionNotesXml,
+                ReadXml);
+
+            // this method does not save the object
+            _saved = false;
```
bhelx: Not sure if this is needed.

Library/Subscription.cs
cbarton: Ya, I like that idea

Library/Subscription.cs
cbarton: Maybe we can pass in the notes to the `WriteSubscriptionNotesXml` method and then if the save is successful assign them to the subscription

bhelx:
:+1:

PR #74, opened 2015-03-23
Do not parse nil as a boolean for the accounts vat location valid

cbarton:
The parse blows up when parsing a `VatLocationValid` field of type `nil`. This will avoid trying to parse in such a case and allow that tri-state field to exist.

cc/ @bhelx 

tests:

  - with VAT location validation turned on and an account that is not in the EU
  - The following should not blow up and return `False`: 
```
var account = Accounts.get('<account_code>');
Console.WriteLine(account.VatLocationValid);
```

  - with an account in the EU (that has been validated)
  - the above snippet should reflect the validation of that account.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,8 @@
 Unreleased
 ==================
 
+* fixed; nil VatLocationValid on Account would throw a parse error
+
 1.1.8 (stable) / 2015-01-26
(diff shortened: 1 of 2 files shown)


bhelx:
:+1:

PR #78, opened 2015-04-21
Added support for Diners Club

brettmas:
This is mainly to fix the exception that is thrown when using Diner's club:
> Requested value "Diner'sClub" was not found

The API seems to return "Diner's Club" as the card type, should really be Diners Club as in their branding.


cbarton:
Thank you for your contribution! Can you add a CHANGELOG entry and squash the commits?

PR #98, opened 2015-08-21
Adding setup_fee_accounting_code

csmb:
Exposing a `setup_fee_accounting` code to the client library since it's now available.

Also, added another link in the Readme to our docs site where folks can find examples of how to make calls with this lib and updated spacing.


bhelx:
:+1:

PR #100, opened 2015-08-31
Add GetRedemptions to Subscription

bhelx:
Trying to fix. Seems to only be reading one element right now.

Approvers: @cbarton

```csharp
var sub = Subscriptions.Get("308d1f45319d5ae8cb15b742f194114c");
            
var redemptions = sub.GetRedemptions();

Console.WriteLine(redemptions);
Console.WriteLine(redemptions.Count);
Console.WriteLine(redemptions[0].AccountCode);
```

### DIFF
--- a/Library/Subscription.cs
+++ b/Library/Subscription.cs
@@ -374,6 +374,16 @@ namespace Recurly
             return true;
         }
 
+        public RecurlyList<CouponRedemption> GetRedemptions()
+        {
+            var coupons = new CouponRedemptionList();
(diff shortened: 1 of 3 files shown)


Library/Subscription.cs
```diff
@@ -374,6 +374,16 @@ public bool UpdateNotes(Dictionary<string, string> notes)
             return true;
         }
 
+        public RecurlyList<CouponRedemption> GetRedemptions()
```
bhelx: I wonder if we need to change this method name to `GetActiveRedemptions` like the account object

PR #102, opened 2015-09-01
Coupons 2.0 changes

bhelx:
These are some of the changes for coupons. To be merged before these two:

* https://github.com/recurly/recurly-client-net/pull/100
* https://github.com/recurly/recurly-client-net/pull/101

Approvers: @shulmang

### DIFF
--- a/Library/CouponRedemption.cs
+++ b/Library/CouponRedemption.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Xml;
 
 namespace Recurly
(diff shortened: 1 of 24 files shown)


Library/CouponRedemption.cs
```diff
@@ -9,6 +9,7 @@ namespace Recurly
     public class CouponRedemption : RecurlyEntity
     {
 
+        public string UUID { get; private set; }
```
bhelx: @shulmang Looking at this again with this lib fresh on my mind, what would you think about me changing this to `Uuid` to be consistent with the other instance variables?

shulmang: @bhelx maintaining consistency makes sense to me.

shulmang:
:+1:

bhelx:
Should be published in 1.2.4 https://www.nuget.org/packages/recurly-api-client/

PR #103, opened 2015-09-15
Fix AddOn.GetHashCode() exception

### LINKED ISSUE #75: Getting AddOns from Plan -> Addons doesn't populate PlanCode for add on
Consider the following pseudo code.

var plan = Plans.Get("some plan code")
plan.AddOns[0].GetHashCode() //Exception

The problem is that when the AddOns collection is fetched the PlanCode property is not set in the ReadXml method of the AddOn class. The AddOn class overrides GetHashCode which just does `return PlanCode.GetHashCode();` Since the property is null GetHashCode naturally fails.

https://github.com/recurly/recurly-client-net/blob/master/Library/AddOn.cs#L102

I discovered this as I'm using Automapper and it uses GetHashCode internally.

MuffinMan10110:
Primarily fixes #75 

Also fixed the following unit tests:
 - Create plan passed invalid characters to `SetupFeeAccountingCode`
  - The Plan WriteXML also incorrectly used the value for `AccountingCode` (instead of `SetupFeeAccountingCode`) when writing `setup_fee_accounting_code`
 - GetActiveRedemption returns null instead of throwing an exception so I check for null instead

### DIFF
--- a/Library/AddOn.cs
+++ b/Library/AddOn.cs
@@ -111,7 +111,6 @@ namespace Recurly
 
                 switch (reader.Name)
                 {
-
                     case "add_on_code":
                         AddOnCode = reader.ReadElementContentAsString();
                         break;
@@ -139,7 +138,6 @@ namespace Recurly
                     case "tax_code":
                         TaxCode = reader.ReadElementContentAsString();
                         break;
-
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


bhelx:
:+1:

bhelx:
Should be published in 1.2.4 https://www.nuget.org/packages/recurly-api-client/

Issue #110, opened 2015-10-07
Get Tests and CI running


bhelx:
In the rest of our libraries, we use travis to run the tests and don't merge anything where the tests do not pass. We should be doing the same thing in this library. In order to do this, we should add a travis config and fix the tests (or remove the ones that aren't working). Once we have travis protecting PRs we can get more serious about adding tests for each change.

@MuffinMan10110 let me know if you have any input on this. I plan to get this done in the next few weeks.

MuffinMan10110:
@bhelx Your idea sounds good to me. 

Note that a bit of work is going to be required to get the tests in a place to be run in an automated environment.

1. I know there are a handful of tests that are either deprecated or are impossible to run in the current state of the Recurly account associated with the library.
2. Last time I checked some tests encountered one or more race conditions and failed when being run with all of the others, retrying them one at a time was successful.

bhelx:
@MuffinMan10110 I noticed the same.

My plan was to just delete (or maybe comment out) any tests that are causing problems. I'll attempt to fix them but if I can't find a good fix or I don't understand them, they will be removed. I'm definitely more concerned about getting `some` tests running at all than having `all` of them running. We can work on incrementally improving them from that point forward. We can just draw a line in the sand at that point. It's not ideal but I think it's the best way to get started on making things better.

MuffinMan10110:
@bhelx That makes sense about getting some tests running.

bhelx:
Closing with https://github.com/recurly/recurly-client-net/pull/113

I think some more improvements could be made in the future but we'll open up specific tickets for that

bhelx:
Just a note, if we want to put integration tests into travis in the future we'll need to find out what is making them so inconsistent as well as slow.

PR #113, opened 2015-10-07
Add Travis file and skip failing test

bhelx:
Goal is to get Travis CI working as described here: https://github.com/recurly/recurly-client-net/issues/110

My idea is to replace all FactAttributes with our own `RecurlyFact` in which `scenarios` can be added. When run in travis, the tests will skip all the `Integration` scenarios.

This skips a lot of tests but at least gets something running. We can start enforcing unit tests after this.
Also, I can run the integration tests locally before releases and before merging a PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: csharp
+solution: Recurly.sln
+
+install:
+  - nuget restore Recurly.sln
+  - nuget install xunit.runners -Version 1.9.2 -OutputDirectory testrunner
+script:
+  - xbuild /p:Configuration=Release Recurly.sln
+  - mono ./testrunner/xunit.runners.1.9.2/tools/xunit.console.clr4.exe ./Test/bin/Release/Recurly.Test.dll
--- a/Test/AccountTest.cs
+++ b/Test/AccountTest.cs
@@ -6,7 +6,7 @@ namespace Recurly.Test
 {
     public class AccountTest : BaseTest
     {
-        [Fact]
+        [RecurlyFact(TestEnvironment.Type.Integration)]
         public void CreateAccount()
         {
             var acct = new Account(GetUniqueAccountCode());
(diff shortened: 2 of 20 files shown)


bhelx:
@MuffinMan10110 perhaps you have some opinions on this?

MuffinMan10110:
So what you're suggesting is skip the integration tests in Travis until we can fix them so that running them in quick succession won't break them?

bhelx:
@MuffinMan10110 If you run all the tests locally now they all pass. The idea is to not run integration tests in travis because they take too long and I don't want travis pounding our server all day. I'd prefer to start writing unit tests moving forward. We can leave running integration tests as a local thing and I can run them when appropriate.

MuffinMan10110:
@bhelx I hear what you're saying about not wanting to run the integration tests in your CI workflow and I think your compromised approach is great. It's a similar approach that I've taken in the past.

On the other hand I think there's a lot of value in working towards completely automating this solution. Essentially upon approval of a pull request Travis could build and run all tests and if successful merge into master which would then automatically publish to Nuget. Your job is now way easier, less error prone and everyone gets updates essentially as fast as you can get around to reviewing the pull request.

The concerns you bring up are definitely valid, but can be worked around:
- Tests can be run in parallel on quad/hex+ core CPU
- Recurly API server can be scaled out or hardware beefed up. Cloud services allow you to scale up/down automatically.

If you're interested in taking the automation further in the future I would be glad to help, but what you've proposed about just running the integration tests locally sounds fine to me. :+1:

bhelx:
@MuffinMan10110 good points. I'm all for more automation. I think we will get there in time but I think this is a good foundation for now. We can always add or change things piecemeal.

csmb:
Cool, great addition! These changes certainly increase confidence in adding or otherwise altering this client.

PR #115, opened 2015-10-28
adds ability to get invoice from subscription

cgerrior:
Currently it is only possible to get the linked Account from a Subscription object. This PR provides an .Invoice method on the Subscription object that allows you to retrieve the linked Invoice.

### DIFF
--- a/Library/Subscription.cs
+++ b/Library/Subscription.cs
@@ -56,6 +56,16 @@ namespace Recurly
         {
             get { return _account ?? (_account = Accounts.Get(_accountCode)); }
         }
+        
+        private string _invoiceNumber;
+        private Invoice _invoice;
+        /// <summary>
+        /// Invoice in Recurly
+        /// </summary>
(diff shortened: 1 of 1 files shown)


bhelx:
I'll add this to the History when i create the version bump PR.

csmb:
This doesn't build because of the switch statement contains multiple cases with the label value `invoice`.
https://github.com/recurly/recurly-client-net/pull/115/files#diff-4d8eb83e3f91427976dbfd70e1f9691fR552

I believe that if you remove that second instance of the `invoice` label, it should build and we can merge it in!

bhelx:
@cgerrior can you fix? Also could you rebase so we can have the tests run?

bhelx:
@cgerrior nevermind, apparently @csmb has it

csmb:
I'm going to close this out, since I've opened a new PR, here: https://github.com/recurly/recurly-client-net/pull/123

PR #120, opened 2015-11-03
Coupons should default to SingleCode

bhelx:
Coupons should default to Type SingleCode unless specified otherwise. This fixes the tests as well.


csmb:
+1

PR #126, opened 2015-11-17
Fix integration test and add open state

bhelx:
PreviewSubscription() integration test was failing due to not knowing about the Open state as well as thinking the subscription would be active.


csmb:
Looks good to me

PR #129, opened 2015-11-18
Retargeting .NET v4.5

bhelx:
* Retargeting .net v4.5 so we can use TLS headers.
* Adding explicit TLS 1.2 and 1.1 settings.


csmb:
Solid gold! +1 from me

Issue #140, opened 2016-03-18
Update Plan Addon - throw exception


zjazd:
Example from https://dev.recurly.com/docs/update-an-add-on

var plan = Plans.Get("gold");
var addon = plan.GetAddOn("ipaddresses");
addon.UnitAmountInCents["USD"] = 200;
addon.Update();

throws exception beaceuse PlanCode is null, and there is no public method or property that set PlanCode (only internal constructor)

bhelx:
@zjazd This appears to be a bug. I have a longer term fix in mind but for now I have this PR that should fix the problem:

https://github.com/recurly/recurly-client-net/pull/141

bhelx:
Closed with https://github.com/recurly/recurly-client-net/pull/141

PR #141, opened 2016-03-18
Set AddOn#PlanCode for GetAddOn

bhelx:
This is to address https://github.com/recurly/recurly-client-net/issues/140

This ensures that PlanCode can be publicly set and will be set when an addon is fetched so it can be updated. 

Ideally we would not expose the `PlanCode` setter here. This is something I can find a workaround for or something i can revert in the follow up PR.

We really should do a follow up where we change the Update() methods to use the resource's hrefs (which we should also be reading from xml) that way the resources don't have to track all of their parent resources and we don't need to encode routing logic into the objects. I'm going to do a follow up PR for that https://github.com/recurly/recurly-client-net/issues/142

### DIFF
--- a/Library/AddOn.cs
+++ b/Library/AddOn.cs
@@ -7,7 +7,7 @@ namespace Recurly
     public class AddOn : RecurlyEntity
     {
 
-        public string PlanCode { get; private set; }
+        public string PlanCode { get; set; }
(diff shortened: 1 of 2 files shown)


csmb:
Looks good to me +1

PR #143, opened 2016-03-22
Fix TLS preferences for releases

bhelx:
The release build did not contain the line of code with TLS preferences due to these constants not being present. Was discovered by a report from @rafdizzle86 and decompiling the nuget package :/

In order to test, build a release then decompile and check that this line is present: https://github.com/recurly/recurly-client-net/blob/master/Library/Client.cs#L37

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: csharp
+dist: trusty
 mono:
-  - weekly
+  - 4.0.5
 solution: Recurly.sln
-
 install:
   - nuget restore Recurly.sln
   - nuget install xunit.runners -Version 1.9.2 -OutputDirectory testrunner
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: csharp
+mono:
+  - weekly
 solution: Recurly.sln
 
 install:
--- a/Library/Client.cs
+++ b/Library/Client.cs
@@ -30,12 +30,7 @@ namespace Recurly
         protected Client(Settings settings)
         {
             Settings = settings;
-           
-            // Don't include this on Travis build
-            // .NET 4.0 does not have these constants
-#if NOT_RUNNING_ON_4_0
             ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Tls11;
-#endif
         }
 
         internal static void ChangeInstance(Client client)
--- a/Library/Client.cs
+++ b/Library/Client.cs
@@ -30,7 +30,12 @@ namespace Recurly
         protected Client(Settings settings)
         {
             Settings = settings;
-            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Tls11;
+
+            // SecurityProtocolType values below not available in Mono < 4.3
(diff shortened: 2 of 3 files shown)


bhelx:
This is breaking in travis now. Will need to investigate a better way to do this.

MuffinMan10110:
I notice that a condition was added to only include the TLS preferences if the Target Framework wasn't 4.0, but I believe that is N/A because the library only targets .NET 4.5

https://github.com/recurly/recurly-client-net/blob/master/Library/Recurly.csproj#L13

MuffinMan10110:
Also @bhelx it appears that Travis doesn't support .NET 4.5 because Mono only supports up to 4.0. I did not realize this back when we were discussing targeting 4.5.

bhelx:
@MuffinMan10110 It should be fine as long as we can get a constant at compile time to allow us to skip this line https://github.com/recurly/recurly-client-net/blob/master/Library/Client.cs#L37

 I'm wondering if there is a way to define a constant based on the environment variables?

Perhaps maybe there is a way to make a propertygroup specific to travis?
https://github.com/recurly/recurly-client-net/pull/143/files#diff-c3879b3ee73000365a2fff28d525723dR36

MuffinMan10110:
@bhelx I see the original SO question that those constants were referenced from and if you look at the second answer they explain the flaw in the accepted answer http://stackoverflow.com/questions/3436526/detect-target-framework-version-at-compile-time

Will have time to take a look at this closer tomorrow.

MuffinMan10110:
Also, no reason have those two contstants.

You could just negate the first:
      
     #IF !NOT_RUNNING_ON_4

bhelx:
Quick update on this. I have changed it to remove this conditional since these constants are supposed to be in Mono now. I think this is an issue with travis not using the most up to date Mono:

https://github.com/travis-ci/travis-ci/issues/5219

bhelx:
I'm going to attempt to change the mono channel to `weekly` as suggested.

bhelx:
I got this to work by copying the Kestrel project https://github.com/aspnet/KestrelHttpServer/blob/14bbba06418234f1f6d41eb2540ca26f8598fd23/test/Microsoft.AspNetCore.Server.KestrelTests/HttpsConnectionFilterTests.cs#L30

I will need to verify it's still working on an MS .NET build

bhelx:
@MuffinMan10110 any opinions?

MuffinMan10110:
That's pretty clever and I never thought of that. It appears that you can assign an enum variable using a value that isn't in the enum because it's just backed by an int.

Here's proof that it works: https://dotnetfiddle.net/qJzYSS
 
And here's the MS definition of that Enum.

```
namespace System.Net
{
  /// <summary>
  /// Specifies the security protocols that are supported by the Schannel security package.
  /// </summary>
  [System.Flags]
  public enum SecurityProtocolType
  {
    Ssl3 = 48,
    Tls = 192,
    Tls11 = 768,
    Tls12 = 3072,
  }
}
```

bhelx:
@MuffinMan10110 yeah, enums in most languages are just named ints. So you can cast them.

bhelx:
@csmb let's get this merged

csmb:
+1

bhelx:
woo

PR #150, opened 2016-05-12
Fixed Get Adjustment and related test

### LINKED ISSUE #149: Adjustments.Get returns uninitialized object
I am using version 1.3.1.
When I call this code (I don't see in you public Library that this code has changed):

```
    public class Adjustments
    {
        public static Adjustment Get(string uuid)
        {
            var adjustment = new Adjustment();
            Client.Instance.PerformRequest(Client.HttpRequestMethod.Get,
                "/adjustments/" + Uri.EscapeUriString(uuid),
                adjustment.WriteXml);
            return adjustment;
        }
    }

```

the returned object is not initialized. Also there is no check on the return code on the PerformRequest.  It seems this should use ReadXml instead?? and also check the return code like:

```
        public static Invoice Get(string invoiceNumberWithPrefix)
        {
            var invoice = new Invoice();
            
            var statusCode = Client.Instance.PerformRequest(Client.HttpRequestMethod.Get,
                Invoice.UrlPrefix + invoiceNumberWithPrefix,
                invoice.ReadXml);

            return statusCode == HttpStatusCode.NotFound ? null : invoice;
        }

```

MuffinMan10110:
fixes #149

### DIFF
--- a/Library/Coupon.cs
+++ b/Library/Coupon.cs
@@ -71,7 +71,7 @@ namespace Recurly
         public CouponDiscountType DiscountType { get; private set; }
         public CouponState State { get; private set; }
         public RedemptionResourceType RedemptionResource { get; set; }
-        public CouponType Type { get; set; }
+        public CouponType Type { get; set; } = CouponType.SingleCode;
 
         /// <summary>
         /// A dictionary of currencies and discounts
--- a/Library/Coupon.cs
+++ b/Library/Coupon.cs
@@ -71,7 +71,7 @@ namespace Recurly
         public CouponDiscountType DiscountType { get; private set; }
         public CouponState State { get; private set; }
         public RedemptionResourceType RedemptionResource { get; set; }
-        public CouponType Type { get; set; } = CouponType.SingleCode;
+        public CouponType Type { get; set; }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


bhelx:
Thanks @MuffinMan10110 ! I've been out of town. Should get this merged today or tomorrow.

Library/Coupon.cs
```diff
@@ -71,7 +71,7 @@ public enum CouponType
         public CouponDiscountType DiscountType { get; private set; }
         public CouponState State { get; private set; }
         public RedemptionResourceType RedemptionResource { get; set; }
-        public CouponType Type { get; set; } = CouponType.SingleCode;
```
bhelx: @MuffinMan10110 what is the purpose of this change and what are it's implications?

MuffinMan10110: @bhelx I didn't realize that is C# 6.0 syntax for setting the default value for a property. It didn't compile for me and I thought I had accidentally put it there so I removed it. My change should be reverted.

MuffinMan10110:
@bhelx I reverted the change in question back to what it was.

bhelx:
Thanks @MuffinMan10110 !

PR #154, opened 2016-07-19
Update to API version 2.2

bhelx:
Updating to API version 2.2

I had to add support for `updated_at` in the appropriate places. I'd like to also add usage based billing but I think that might take some more time and I want to assess the demand. This is good enough to unblock us and get us to version 2.3.

### DIFF
--- a/Library/AddOn.cs
+++ b/Library/AddOn.cs
@@ -6,11 +6,16 @@ namespace Recurly
 {
     public class AddOn : RecurlyEntity
     {
-
         public string PlanCode { get; set; }
         public string AddOnCode { get; set; }
         public string Name { get; set; }
-
+        public int DefaultQuantity { get; set; }
+        public bool? DisplayQuantityOnHostedPage { get; set; }
+        public string TaxCode { get; set; }
+        public bool? Optional { get; set; }
+        public string AccountingCode { get; set; }
+        public DateTime CreatedAt { get; private set; }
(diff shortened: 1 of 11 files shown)


Library/AddOn.cs
```diff
@@ -6,11 +6,16 @@ namespace Recurly
...
-
         public string PlanCode { get; set; }
         public string AddOnCode { get; set; }
         public string Name { get; set; }
-
+        public int DefaultQuantity { get; set; }
+        public bool? DisplayQuantityOnHostedPage { get; set; }
+        public string TaxCode { get; set; }
+        public bool? Optional { get; set; }
+        public String AccountingCode { get; set; }
```
csmb: Should we add `xmlWriter.WriteElementString("accounting_code", AccountingCode);` [further down](https://github.com/recurly/recurly-client-net/pull/154/files#diff-e84c4a9c2623e09a15d1c2b70fecc921R164)?

bhelx: Yes, appears I missed it. Good eye.

MuffinMan10110: @csmb Looks like the API accepts setting that value when creating a plan add-on. If you don't set it then it gets set to the value of `AddOnCode`. So it should be a valid field to write to the request xml.

bhelx:
@csmb okay updated the reader and writer. Sorry it took so long.

csmb:
+1

PR #162, opened 2016-09-09
WIP: Update to config with API Access

BenjaminGibbs:
I have updated the settings component so it now does not load the settings by default, and allows you to specify how you load the config. I am looking for feedback if this is the direction being thought of and how it could be adjusted to fit more inline with your library.

The use case would now be inline with other libraries work where you either tell it to load from config, or inject the settings via parameters..

Usage would be to call one of the following when the app starts up.

```c#
Recurly.Configuration.SettingsManager.Initialize("api", "subdomain", "private", 100);

-or-

Recurly.Configuration.SettingsManager.InitializeFromConfig()
```

This would be a breaking change as it doesn't load from the config file by default in the current state, it may be preferred to try and load that so this is not breaking.

### DIFF
--- a/Library/Configuration/Settings.cs
+++ b/Library/Configuration/Settings.cs
@@ -52,6 +52,7 @@ namespace Recurly.Configuration
             get { return _instance ?? (_instance = new Settings()); }
         }
 
+
         public void InitializeFromConfig()
         {
             ApiKey = Section.Current.ApiKey;
@@ -68,13 +69,5 @@ namespace Recurly.Configuration
             PageSize = pageSize;
         }
 
-
-        //private Settings()
-        //{
-        //}
-
-        //internal Settings(string apiKey, string subdomain, string privateKey, int pageSize)
-        //{
-        //}
     }
 }
--- a/Library/Configuration/Settings.cs
+++ b/Library/Configuration/Settings.cs
@@ -10,10 +10,57 @@ namespace Recurly.Configuration
     internal class Settings
     {
         // non-static, as these change per instance, likely
-        public string ApiKey { get; private set; }
-        public string PrivateKey { get; private set; }
-        public string Subdomain { get; private set; }
-        public int PageSize { get; private set; }
+        public string ApiKey
+        {
+            get
+            {
+                if (_hasLoaded == false)
+                {
+                    throw new Exception("The Recurly client has has no configuration initialized, please add your settings to web/app.config or call Recurly.Configuration.SettingsManager.Initialize(args)");
+                }
+                return _apiKey;
+            }
+            private set { _apiKey = value; }
+        }
+
+        public string PrivateKey
+        {
+            get
+            {
+                if (_hasLoaded == false)
+                {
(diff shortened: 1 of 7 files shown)


bhelx:
This is pretty great. Thank you! I'll be looking at this as soon as I get the chance tomorrow.

bhelx:
@MuffinMan10110 do you have any feedback on this?

@BenjaminGibbs is there any way to have it continue working for our current setup but print out a deprecation message (or maybe use the Obsolete attribute) ? We try not to make hard breaking changes.

MuffinMan10110:
@BenjaminGibbs Thanks for making this update, it will help some people who are using things like Azure to specify their credentials.

I agree with @bhelx that it should not be breaking existing integrations.

I suggest that by default the configuration continues to be loaded from the web/app.config, however your changes would allow overriding it if the web.config section already exists, or of course if the web.config section doesn't exist one would need to specify the credentials using the new way.

Additionally (if this check isn't already in place), if settings weren't set at all the library should throw a meaningful exception. I imagine it would be cryptic if a null reference exception was thrown when trying to use the library for the first time.

Would like to hear your feedback on the suggestions provided above.

BenjaminGibbs:
So we could add the constructor back in to the Settings class, and call the `InitializeFromConfig()` in a try block. That would have the desired effect of not breaking current integrations.

We could then have a private var within that class let's say `hasLoaded` and set it to true once the data is loaded via the config or the api call. If a user tries to access any of the 4 properties `ApiKey, PrivateKey, Subdomain or PageSize` and `hasLoaded==false` we throw a friendly exception telling them that they need to load a config.

If we did the above we could then remove `Recurly.Configuration.SettingsManager.InitializeFromConfig()` as in theory the config file is loaded by default, unless there is some use case where someone may wish to recall this but I am not sure I can think of one.

How would that work @MuffinMan10110 and @bhelx ?

MuffinMan10110:
@BenjaminGibbs What you described sounds like a good approach to me.

BenjaminGibbs:
I have updated.

- It now loads from the .config files by default, but should slightly fail if it can't.
- If you try and access the settings from the settings class and settings have not yet been loaded then it will throw an exception.
- I have left in the `InitializeFromConfig()` in the settings manager for now, while the use case for calling it is very edge case if we remove it there is no way to reload settings from the config files as the methods which do that are all marked internal. - What are you thoughts on this?
- The class name `SettingsManager`, do you have any better suggestions or happy with it?

bhelx:
@BenjaminGibbs Thanks. Let me try this out and get back to you.

bhelx:
@BenjaminGibbs This seems to have worked for me using my current config setup. Thanks for making it backwards compatible!

@MuffinMan10110 Do you have any other comments?

MuffinMan10110:
@bhelx Looks good, thanks.

bhelx:
@MuffinMan10110 thanks for taking the time to check it out!

@BenjaminGibbs I'm planning on doing a release tomorrow. I'll be adding some instructions to the README reflecting this change. Thanks again for your patience.

BenjaminGibbs:
@bhelx No problems great it has been released, I can start integrating it now.

bhelx:
Thanks, don't hesitate to request any changes or new features.

PR #164, opened 2016-09-14
Add error case for 400 status code

bhelx:
I found this missing case for HTTP status code 400. It's possible this is being misapplied on the server but we should handle it nonetheless.

### DIFF
--- a/Library/Client.cs
+++ b/Library/Client.cs
@@ -203,6 +203,7 @@ namespace Recurly
                         errors = Error.ReadResponseAndParseErrors(response);
                         throw new InvalidCredentialsException(errors);
 
+                    case HttpStatusCode.BadRequest:
                     case HttpStatusCode.PreconditionFailed:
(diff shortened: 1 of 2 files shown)


Library/Client.cs
```diff
@@ -203,6 +203,10 @@ public HttpStatusCode PerformRequest(HttpRequestMethod method, string urlPath)
                         errors = Error.ReadResponseAndParseErrors(response);
                         throw new InvalidCredentialsException(errors);
 
+                    case HttpStatusCode.BadRequest:
```
csmb: [I believe](https://msdn.microsoft.com/en-us/library/06tc147t.aspx) you can move this case label atop `case HttpStatusCode.PreconditionFailed:` on line 210 to since they throw the same exception.

bhelx: Oh, interesting syntax. Thanks!

csmb:
Looks good to me

Issue #168, opened 2016-09-21
Usage Logging crashing when calling Create.


BenjaminGibbs:
So I have ran in to a problem when logging usage. The exception is "{"Input string was not in a correct format."}"

The usage gets logged in the portal just fine but it appears the return message is not getting serialized correctly.

```
   at System.Number.ParseSingle(String value, NumberStyles options, NumberFormatInfo numfmt)
   at System.Xml.XmlConvert.ToSingle(String s)
   at System.Xml.XmlReader.ReadElementContentAsFloat()
   at Recurly.Usage.ReadXml(XmlTextReader reader) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Usage.cs:line 114
   at Recurly.Client.ReadWebResponse(HttpWebResponse response, ReadXmlDelegate readXmlDelegate, ReadXmlListDelegate readXmlListDelegate, ReadResponseDelegate responseDelegate) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Client.cs:line 382
   at Recurly.Client.PerformRequest(HttpRequestMethod method, String urlPath, WriteXmlDelegate writeXmlDelegate, ReadXmlDelegate readXmlDelegate, ReadXmlListDelegate readXmlListDelegate, ReadResponseDelegate reseponseDelegate) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Client.cs:line 170
   at Recurly.Usage.Create() in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Usage.cs:line 56
   at recurlyupdatedtest.MainWindow.GetAccount() in C:\Users\benja\Documents\Visual Studio 2015\Projects\recurlyupdatedtest\recurlyupdatedtest\MainWindow.xaml.vb:line 63
   at recurlyupdatedtest.MainWindow.VB$StateMachine_1_button_Click.MoveNext() in C:\Users\benja\Documents\Visual Studio 2015\Projects\recurlyupdatedtest\recurlyupdatedtest\MainWindow.xaml.vb:line 8
```

The code I am using to get this is:

```vbnet
        Dim demoAccount = Accounts.Get(accountKey)
        Dim subscriptions = demoAccount.GetSubscriptions(Subscription.SubscriptionState.Active)
        Dim firstSub = subscriptions(0)
        Dim usage = New Usage(firstSub.Uuid, "upload")
        usage.Amount = 1
[...]

PR #169, opened 2016-09-21
Do not parse nil usage_percentage

bhelx:
Attempts to fix issue https://github.com/recurly/recurly-client-net/issues/168


csmb:
+1

PR #183, opened 2016-10-26
Allow to skip trial period by setting date in past

skynet2:
Allow to skip trial period by setting date in past refer to https://dev.recurly.com/docs/preview-subscription


bhelx:
:+1: I don't think this should have been here in the first place. Thanks for catching!

PR #186, opened 2016-11-07
allow addresses to be null, and don't force-create one when fetching …

### LINKED ISSUE #179: Serializing empty address
`Account` serializes null addresses. It should not do this:

``` csharp
var  account = new Account("my-account-code"); // existing account, just need code
```

``` xml
<account>
  <account_code>my-account-code</account_code>
  <address>
    <address1/>
    <address2/>
    <city/>
    <!-- ..... -->
  </address>
</account>
```

Donnotron666:
Resolves #179 

Allow addresses to be null, and don't force-create one when fetching it from the account

### DIFF
--- a/Library/Account.cs
+++ b/Library/Account.cs
@@ -42,12 +42,7 @@ namespace Recurly
         public DateTime CreatedAt { get; private set; }
         public DateTime UpdatedAt { get; private set; }
         public bool VatLocationValid { get; private set; }
-
-        public Address Address {
-			get { return _address; }
-            set { _address = value; }
-        }
-        private Address _address;
+        public Address Address { get; set; }
 
         private BillingInfo _billingInfo;
 
@@ -241,7 +236,7 @@ namespace Recurly
         public RecurlyList<Subscription> GetSubscriptions(Subscription.SubscriptionState state = Subscription.SubscriptionState.All)
         {
             return new SubscriptionList(UrlPrefix + Uri.EscapeUriString(AccountCode) + "/subscriptions/"
-                + Build.QueryStringWith(state.Equals(Subscription.SubscriptionState.All) ? "" :  "state=" + state.ToString().EnumNameToTransportCase()));
+                + Build.QueryStringWith(state.Equals(Subscription.SubscriptionState.All) ? "" : "state=" + state.ToString().EnumNameToTransportCase()));
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


Donnotron666:
@bhelx Does this need a version bump?

bhelx:
@Donnotron666 It may, but I general determine those things when I make the release and do that in a separate PR. It depends what else gets into the release.

Test/AccountTest.cs
```diff
@@ -18,18 +18,19 @@ public void CreateAccount()
...
-                Email = "testemail@test.com",
-                FirstName = "Test",
-                LastName = "User",
-                CompanyName = "Test Company",
-                AcceptLanguage = "en",
-                VatNumber = "my-vat-number",
-                TaxExempt = true,
-                EntityUseCode = "I",
-                CcEmails = "cc1@test.com,cc2@test.com"
+			var acct = new Account(GetUniqueAccountCode())
```
bhelx: What's up with the formatting here?

Library/Account.cs
```diff
@@ -44,7 +44,7 @@ public enum AccountState : short
         public bool VatLocationValid { get; private set; }
 
         public Address Address {
-            get { return _address ?? (_address = new Address()); }
+			get { return _address; }
```
bhelx: Strange formatting here. Tabs?

Test/AccountTest.cs
```diff
@@ -29,7 +29,8 @@ public void CreateAccountWithParameters()
                 VatNumber = "my-vat-number",
                 TaxExempt = true,
                 EntityUseCode = "I",
-                CcEmails = "cc1@test.com,cc2@test.com"
+                CcEmails = "cc1@test.com,cc2@test.com",
+                Address = new Address()
```
bhelx: We're actually going to need a public constructor for `Address`. The only reason this works in the test is because the test are `internal` to the library.

bhelx:
So, the issue here is that the current API for creating addresses on an account looks something like this:

```csharp
var account = new Account("asbcd123456");
account.Address.Address1 = "hello";
account.Address.City = "whatever";
// ....
account.Create();
```

The API protects you from having to creates addresses which I feel like is kind of needless and probably makes it harder to map some address data to an address.

As you noted before, changing this to the programmer's responsibility to create the Address breaks the API but I think for the better.

I'll write it up in the release notes and link here.

Library/Account.cs
```diff
@@ -44,7 +44,7 @@ public enum AccountState : short
         public bool VatLocationValid { get; private set; }
 
         public Address Address {
-            get { return _address ?? (_address = new Address()); }
+            get { return _address; }
             set { _address = value; }
```
bhelx: Do we need the `_address` anymore? I think we can remove it and change this line to be:

```csharp
public Address Address { get; set; }
```

Test/AccountTest.cs
bhelx: For reference: https://github.com/recurly/recurly-client-net/blob/master/Library/Address.cs#L15-L20

Library/Account.cs
```diff
@@ -42,12 +42,7 @@ public enum AccountState : short
         public DateTime CreatedAt { get; private set; }
         public DateTime UpdatedAt { get; private set; }
         public bool VatLocationValid { get; private set; }
-
-        public Address Address {
-            get { return _address ?? (_address = new Address()); }
-            set { _address = value; }
-        }
-        private Address _address;
+		public Address Address { get; set; }
```
bhelx: Still formatted weird?

bhelx:
:+1:

PR #188, opened 2016-11-10
Add UsageList class

### LINKED ISSUE #187: Get subscription addon usage
Hi. 
There is no implement for getting subscription addon usage. Is this missed currently or it is somewhere I don't know.

Thanks in advance.
An.

bhelx:
Resolves #187 

```csharp
var subscriptionUuid = "37bfef7a8e44cfc3817b7a43eba8a6e6";
var subscriptionAddOnCode= "ip_addresses";

var usages = Usage.List(subscriptionUuid, subscriptionAddOnCode);
while (usages.HasAny())
{
  foreach (var u in usages)
    Console.WriteLine("Usage: " + u.Amount);
  usages = usages.Next;
}
```

### DIFF
--- a/Library/Recurly.csproj
+++ b/Library/Recurly.csproj
@@ -64,6 +64,7 @@
   <ItemGroup>
     <Compile Include="Account.cs" />
     <Compile Include="Address.cs" />
+    <Compile Include="List\UsageList.cs" />
     <Compile Include="List\ShippingAddressList.cs" />
     <Compile Include="ShippingAddress.cs" />
     <Compile Include="Configuration\SettingsManager.cs" />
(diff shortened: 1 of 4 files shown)


Library/Recurly.csproj
```diff
@@ -64,6 +64,7 @@
   <ItemGroup>
     <Compile Include="Account.cs" />
     <Compile Include="Address.cs" />
+    <Compile Include="List\UsageList.cs" />
```
csmb: Want to move this down with it's friend `Usage.cs` on L:72

bhelx: This is automatically generated, not sure how it works. Would be nice to organize them but it might make a bunch of noise in commits. Will investigate when we get this working without visual studio!

PR #196, opened 2016-12-07
Started adding support for Async methods

MuffinMan10110:
Because of the number of methods and tests to add only a subset of existing methods have an Async counterpart, mainly the methods in Subscription.cs. The library remains backwards compatible because the Async methods are new methods with the "Async" suffix. My goal is, over time to submit additional pull requests to support more endpoints.

Additionally, in order to accomplish this the following related changes were made:

- Fixed a couple of broken unit tests
- Updated all Xunit nuget packages to the latest versions in order to get support for testing async tests. Only a handful of Asserts in the SubscriptionTests had to be updated after the upgrade.


bhelx:
This is great work! Thanks for taking it on! I'd be happy to help out once I understand what is going on. Also, if there are any small bugs you find it would be great to send them in separate PRs so those can get merged quick.

MuffinMan10110:
@bhelx Thanks, I'd love to have the help. It looks like there is an issue with the tests so I will look into that and hopefully get another commit in this weekend.

MuffinMan10110:
@bhelx I think a better approach would be to scrap this pull request and create a new one with just the tests updated since so many files changed in the process. I can then followup with a separate pull request for the async stuff.

Issue #211, opened 2017-01-19
Requested value 'Processing' was not found.


NPSF3000:
We're using this lovely library to process our payments, but occasionally it's throwing an exception when returning the list of Invoices.  We suspect this is because the invoice state is 'Processing' and the library doesn't understand than status.

This bug is quite urgent, as it's breaking live functionality for paying customers.

Thanks!

System.ArgumentException: Requested value 'Processing' was not found.
   at System.Enum.TryParseEnum(Type enumType, String value, Boolean ignoreCase, EnumResult& parseResult)
   at Recurly.StringExtensions.ParseAsEnum[T](String source) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Extensions\StringExtensions.cs:line 54
   at Recurly.Invoice.ReadXml(XmlTextReader reader) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Invoice.cs:line 261
   at Recurly.InvoiceList.ReadXml(XmlTextReader reader) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\List\InvoiceList.cs:line 31
   at Recurly.Client.ReadWebResponse(HttpWebResponse response, ReadXmlDelegate readXmlDelegate, ReadXmlListDelegate readXmlListDelegate, ReadResponseDelegate responseDelegate) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Client.cs:line 378
   at Recurly.Client.PerformRequest(HttpRequestMethod method, String urlPath, WriteXmlDelegate writeXmlDelegate, ReadXmlDelegate readXmlDelegate, ReadXmlListDelegate readXmlListDelegate, ReadResponseDelegate reseponseDelegate) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Client.cs:line 169
   at Recurly.Client.PerformRequest(HttpRequestMethod method, String urlPath, ReadXmlListDelegate readXmlListDelegate) in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\Client.cs:line 120
   at Recurly.RecurlyList`1.GetItems() in C:\Users\ben\Documents\GitHubVisualStudio\recurly-client-net\Library\List\RecurlyList.cs:line 89
[...]

PR #215, opened 2017-01-31
Add missing enum values

bhelx:
Related to #211

I've gone through and checked all the enums to make sure they are up to date with what is on the server models. Hopefully this fixes these errors for a while. Any new enum values will be on a future API version release and old versions of the client should be protected from them.

### DIFF
--- a/Library/BillingInfo.cs
+++ b/Library/BillingInfo.cs
@@ -15,7 +15,12 @@ namespace Recurly
             MasterCard,
             AmericanExpress,
             Discover,
-            JCB
+            JCB,
+            Danokrt,
+            Maestro,
+            Forbrugsforeningen,
+            Laser,
+            Unknown
         }
 
         public enum BankAccountType : short
--- a/Library/Invoice.cs
+++ b/Library/Invoice.cs
@@ -15,7 +15,8 @@ namespace Recurly
             Collected,
             Failed,
             PastDue,
-            Processing
+            Processing,
+            Pending
         }
 
         public enum RefundOrderPriority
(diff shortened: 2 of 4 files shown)


Library/BillingInfo.cs
```diff
@@ -15,7 +15,12 @@ public enum CreditCardType : short
             MasterCard,
             AmericanExpress,
             Discover,
-            JCB
+            JCB,
+            Danokrt,
+            Maestro,
+            Forbrugsforeningen,
+            Laser,
+            Unknown
```
bhelx: A couple of these values actually have spaces in them. I left them out. I wonder if we should change this to a string instead of an enum.

1TheMuffinMan: The name of the value from an API doesn't need to match the enum name. We'll have to sanitize the value somewhere such as the xml reader to remove spaces in the value and then it will match the enum name.

bhelx: Okay, I can do a follow up with that.

bhelx:
Let's wait until tomorrow to merge and release this.

NPSF3000:
Just checking in to see how things are going @bhelx  @1TheMuffinMan ?

1TheMuffinMan:
@NPSF3000 It looks like there's some enums that have spaces in it so we need to make some modifications to the default enum handling to support that. @bhelx may respond sometime today with an update.

bhelx:
@1TheMuffinMan Actually I think we should go ahead and merge as is then do a follow up since these other fields are causing some headaches and I don't think we have anyone on this library using those other credit card values yet. I can do a release immediately after.

1TheMuffinMan:
@bhelx Sounds good to me.

PR #217, opened 2017-02-07
Sorting and Filtering parameters

bhelx:
I'm starting this new branch to implement the new sorting parameters in API version 2.4: https://dev.recurly.com/docs/pagination

I'm looking for some feedback as the interface on this first draft feels pretty awkward. I want to keep backwards compatibility if possible but I'm willing to deprecate some methods.

This is what I did with Java for reference: https://github.com/killbilling/recurly-java-library/pull/142

\cc @1TheMuffinMan 

### Example Use

```csharp
try {
    var filter = FilterCriteria.Instance
        .WithOrder(FilterCriteria.Order.Desc)
        .WithSort(FilterCriteria.Sort.CreatedAt)
        .WithBeginTime(new DateTime(2015, 1, 1))
        .WithEndTime(DateTime.UtcNow);           
    
    var accounts = Accounts.List(Account.AccountState.Active, filter);
    foreach (var account in accounts)
    {
        Console.WriteLine(account.AccountCode);
        Console.WriteLine(account.CreatedAt);
    }

    var subs = Subscriptions.List(Subscription.SubscriptionState.Active, filter);
    foreach (var sub in subs)
    {
        Console.WriteLine(sub.Uuid);
        Console.WriteLine(sub.ActivatedAt);
    }

    var coupons = Coupons.List(Coupon.CouponState.All, filter);
    foreach (var c in coupons)
    {
        Console.WriteLine(c.CouponCode);
        Console.WriteLine(c.CreatedAt);
    }

    var plans = Plans.List(filter);
    foreach (var p in plans)
    {
        Console.WriteLine(p.PlanCode);
        Console.WriteLine(p.CreatedAt);
    }

    var transactions = Transactions.List(TransactionList.TransactionState.All,
        TransactionList.TransactionType.All,
        filter);
    foreach (var t in transactions)
    {
        Console.WriteLine(t.Uuid);
        Console.WriteLine(t.CreatedAt);
    }
}
catch (ValidationException ex)
{
    foreach(Error err in ex.Errors)
    {
        Console.WriteLine(err.ToString());
    }
}
```

### DIFF
--- a/Library/Account.cs
+++ b/Library/Account.cs
@@ -510,7 +510,21 @@ namespace Recurly
         /// <returns></returns>
         public static RecurlyList<Account> List(Account.AccountState state = Account.AccountState.Active)
         {
-            return new AccountList(Account.UrlPrefix + "?state=" + state.ToString().EnumNameToTransportCase());
+            return List(state, null);
+        }
+
+        /// <summary>
+        /// Lists accounts, limited to state
+        /// </summary>
+        /// <param name="state">Account state to retrieve</param>
+        /// <param name="filter">FilterCriteria used to apply server side sorting and filtering</param>
+        /// <returns></returns>
+        public static RecurlyList<Account> List(Account.AccountState state, FilterCriteria filter)
+        {
+            filter = filter.Equals(null) ? FilterCriteria.Instance : filter;
+            var parameters = filter.ToNamedValueCollection();
+            parameters["state"] = state.ToString().EnumNameToTransportCase();
+            return new AccountList(Account.UrlPrefix + "?" + parameters.ToString());
         }
     }
 }
--- a/Library/FilterCriteria.cs
+++ b/Library/FilterCriteria.cs
@@ -1,0 +1,67 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Recurly
+{
+    public class FilterCriteria
+    {
+        public enum Sort
+        {
+            CreatedAt,
+            UpdatedAt
+        }
+
+        public enum Order
+        {
+            Asc,
+            Desc
+        }
+
+        public Sort? SortField;
+        public Order? OrderField;
+        public DateTime? BeginTime;
+        public DateTime? EndTime;
+
+        private FilterCriteria() { }
+
+        public static FilterCriteria Instance { get { return new FilterCriteria(); } }
+
+        public NameValueCollection ToNamedValueCollection()
+        {
+            var nvc = System.Web.HttpUtility.ParseQueryString(string.Empty);
+            if (SortField.HasValue) nvc["sort"] = SortField.Value.ToString().EnumNameToTransportCase();
+            if (OrderField.HasValue) nvc["order"] = OrderField.Value.ToString().EnumNameToTransportCase();
+            if (BeginTime.HasValue) nvc["begin_time"] = BeginTime.Value.ToString("s");
+            if (EndTime.HasValue) nvc["end_time"] = EndTime.Value.ToString("s");
+            return nvc;
+        }
+
+        public FilterCriteria WithSort(Sort sort)
+        {
+            SortField = sort;
+            return this;
+        }
+
+        public FilterCriteria WithOrder(Order order)
+        {
+            OrderField = order;
+            return this;
+        }
+
+        public FilterCriteria WithBeginTime(DateTime begin)
+        {
+            BeginTime = begin;
+            return this;
+        }
+
+        public FilterCriteria WithEndTime(DateTime end)
+        {
+            EndTime = end;
+            return this;
+        }
+    }
+}
--- a/Library/Coupon.cs
+++ b/Library/Coupon.cs
@@ -541,7 +541,24 @@ namespace Recurly
         /// <returns></returns>
         public static RecurlyList<Coupon> List(Coupon.CouponState state = Coupon.CouponState.All)
         {
-            return new CouponList(Coupon.UrlPrefix + (state != Coupon.CouponState.All ? "?state=" + state.ToString().EnumNameToTransportCase() : ""));
+            return List(state, null);
+        }
+
+        /// <summary>
+        /// Lists coupons, limited to state and filtered by given criteria
+        /// </summary>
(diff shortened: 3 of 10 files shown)


Library/Account.cs
```diff
@@ -463,6 +463,7 @@ public override int GetHashCode()
     public sealed class Accounts
     {
         internal const string UrlPrefix = "/accounts/";
+        private static readonly QueryStringBuilder Build = new QueryStringBuilder();
```
bhelx: Can ignore this. Might get rid of this builder in favor of using `NamedValueCollection`

Library/SortParams.cs
```diff
@@ -0,0 +1,39 @@
...
+            UpdatedAt
+        }
+
+        public enum OrderDirection
+        {
+            Asc,
+            Desc
+        }
+
+        public SortField? Sort;
```
bhelx: I think it might make more sense to switch these names around:

```csharp
public Sort? SortField;
```

Library/SortParams.cs
```diff
@@ -0,0 +1,39 @@
...
+        public OrderDirection? Order;
```
bhelx: Same thing here:

```csharp
public Order? OrderDirection;
```

Library/SortParams.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Recurly
+{
+    public class SortParams
```
bhelx: I don't think this is the best name for the class. And maybe it should be namespaced somewhere else.

Library/Account.cs
```diff
@@ -510,7 +511,21 @@ public static void Reopen(string accountCode)
...
+        /// </summary>
+        /// <param name="state">Account state to retrieve</param>
+        /// <param name="sort">SortParams used to describe server sorting</param>
+        /// <returns></returns>
+        public static RecurlyList<Account> List(Account.AccountState state, SortParams sort)
+        {
+            sort = sort.Equals(null) ? new SortParams() : sort;
+            var parameters = sort.toNamedValueCollection();
+            parameters["state"] = state.ToString().EnumNameToTransportCase();
+            return new AccountList(Account.UrlPrefix + "?" + parameters.ToString());
```
bhelx: I may also add a constructor to the `RecurlyList` abstract class to take a base url and a `NamedValueCollection` that way we don't have to keep writing this code.

Library/SortParams.cs
1TheMuffinMan: Another possibility is just to name the field the same as the type if that's more descriptive:
    
    public SortField? SortField;
    public SortDirection? SortDirection

1TheMuffinMan:
If you want to keep backwards compatibility perhaps you can make overloads?

If you're looking to make the interface more straight forward possibly you could implement a fluent interface?

```
var sort = SortParams.Instance
.WithOrder(SortParams.OrderDirection.Desc)
.WithSort(SortParams.SortField.CreatedAt)
.WithEndTime(DateTime.UtcNow);
Accounts.List(Account.AccountState.Active, sort)
```

bhelx:
@1TheMuffinMan I like the idea of chaining the calls. As for the method signatures, I'm thinking of just having two methods for each List. One that takes a sort and one that doesn't.

1TheMuffinMan:
I was thinking the same thing regarding the overloads, keep the existing signature and move the implementation to a new signature that has the sort param and then the old sig. calls the new one.

bhelx:
@1TheMuffinMan Does this look like a proper implementation of a fluent interface? It doesnt seem to be the same as what you have listed. Here is the use:

```csharp
var filter = new FilterCriteria();

filter.Set.WithOrder(FilterCriteria.Order.Desc)
  .WithSort(FilterCriteria.Sort.CreatedAt)
  .WithBeginTime(new DateTime(2015, 1, 1))
  .WithEndTime(DateTime.UtcNow);           
                
var accounts = Accounts.List(Account.AccountState.Active, filter);
```

Could you recommend some improvements or maybe a code sample or article I could read?

1TheMuffinMan:
class Program
    {
        static void Main(string[] args)
        {
            var fc = FilterCriteria.Instance
                .WithOrder(FilterCriteria.OrderDirection.Desc)
                .WithSort(FilterCriteria.SortField.CreatedAt)
                .WithEndTime(DateTime.UtcNow);

            Console.WriteLine(fc.Order);
            Console.WriteLine(fc.Sort);
            Console.WriteLine(fc.EndDate);
        }
    }

    public class FilterCriteria
    {
        private FilterCriteria() { }
        public static FilterCriteria Instance { get { return new FilterCriteria(); } }

        public string MyDate { get; set; }
        public DateTime EndDate { get; private set; }
        public DateTime StartDate { get; private set; }
        public OrderDirection Order { get; private set; }
        public SortField Sort { get; private set; }

        public FilterCriteria WithOrder(OrderDirection order)
        {
            Order = order;
            return this;
        }

        public FilterCriteria WithSort(SortField sort)
        {
            Sort = sort;
            return this;
        }

        public FilterCriteria WithEndTime(DateTime end)
        {
            EndDate = end;
            return this;
        }

        public enum OrderDirection
        {
            Asc,
            Desc
        }

        public enum SortField
        {
            CreatedAt
        }
    }

bhelx:
@1TheMuffinMan thanks for the detailed explanation!

Library/FilterCriteria.cs
```diff
@@ -0,0 +1,65 @@
...
+            Asc,
+            Desc
+        }
+
+        public Sort? SortField;
+        public Order? OrderField;
+        public DateTime? BeginTime;
+        public DateTime? EndTime;
+
+        public static FilterCriteria Instance { get { return new FilterCriteria(); } }
```
1TheMuffinMan: I would add a private or internal instance constructor in addition to the static `Instance` property. That way there wouldn't be any confusion on the preferred way to create the FilterCriteria.

bhelx:
@1TheMuffinMan Okay going to do that today

bhelx:
@1TheMuffinMan made the FilterCriteria constructor private and squashed all commits into 1

1TheMuffinMan:
@bhelx Looks good!

PR #218, opened 2017-02-08
Improvements in UsageList following the same pattern with a public sealed class

jvalladolid:
* Create a new public sealed class to keep the methods to get a list of usages according to the available filters and to get the info of an specific usage record (UsageBillingState and UsageDateTimeType).
* Get the usageId when reading a usage record, to be able to query usage records by usageId.


jvalladolid:
Abandoned

PR #219, opened 2017-02-09
Subscription Change Objects

### LINKED ISSUE #157: BLOCKER wrong request are send on change subscription
``` csharp
var account = Accounts.Get(accountName);
var subs = account.GetSubscriptions(Subscription.SubscriptionState.All);
var s = subs[0];
var newPlan = Plans.Get(planCode);

s.Plan = newPlan;
s.ChangeSubscription(Subscription.ChangeTimeframe.Now);
```

This code should change subscription it does change subscription but charges for 0.

You have to set manualy new plan amount in cents, otherwise it result in corrupted  subscription in recurly.

you have to put this line before change subscription

``` csharp
s.UnitAmountInCents = newPlan.UnitAmountInCents["EUR"];
```

It is antyintuitive, it is against all tutorials and helps, where you should set UnitAmountInCents only if you want to override plan default price.

bhelx:
Resolves #157 

First I will explain the problem. When doing a subscription change, you only want to send across the fields that were modified on the subscription object. In this library, the `Entities` do not track changes on their fields and instead serialize every field that is set. So, suppose you want to only change the plan and you do the `PreviewChange` request. The xml will look like this:

```xml
<subscription>
  <timeframe>now</timeframe> <!-- should be here -->
  <quantity>1</quantity> <!-- should NOT be here -->
  <plan_code>my_new_plan_code</plan_code> <!-- should be here -->
  <subscription_add_ons /> <!-- should NOT be here -->
  <unit_amount_in_cents>0</unit_amount_in_cents> <!-- should NOT be here -->
  <collection_method>automatic</collection_method> <!-- should NOT be here -->
</subscription>
```

So, even though we only wanted to change the plan, we are sending over other fields. So we need to either:
  1. Allow state tracking on entities (can get very messy)
  2. Change the interface such that the programmer applies changes to a fresh object.

I'm attempting the latter for now. Instead of doing this:

```csharp
var subscription = Subscriptions.Get("3b25a12cd76f4a3280ffe54957a7ec59");
subscription.Plan = Plans.Get("gold");  //changing existing plan to gold
var changedSubPreview = subscription.PreviewChange(Subscription.ChangeTimeframe.Now);
```

We'd ask the programmer to create some kind of change object and apply the changes to it instead of the original subscription:

```csharp
var subscription = Subscriptions.Get("3b6aa62f4bac3e177ebefc421fbef29c");
var newPlan = Plans.Get("gold");

// create a new subscription object and set it with the changes you want
var newSubscription = subscription.GenerateChange();
newSubscription.Plan = newPlan;
// if you wanted to make other changes here you could
// subChanges.Quantity = 3;

// now run the change on this fresh subscription object
var subscriptionPreview = newSubscription.PreviewChange(Subscription.ChangeTimeframe.Now);
// you could also call the change and it will update `newSubscription`
newSubscription.ChangeSubscription(Subscription.ChangeTimeframe.Now);
```


bhelx:
@1TheMuffinMan I've updated this code to what I think is a reasonable way to go about it for now. This is ready for some review. The main thing I'm concerned about right now is whether or not the interface is too confusing. After we merge it I can update the documentation to reflect this new way of doing changes and previews.

Library/Subscription.cs
```diff
@@ -801,10 +815,15 @@ protected void WriteChangeSubscriptionXml(XmlTextWriter xmlWriter, ChangeTimefra
...
 
             xmlWriter.WriteElementString("timeframe", timeframe.ToString().EnumNameToTransportCase());
-            xmlWriter.WriteElementString("quantity", Quantity.AsString());
             xmlWriter.WriteStringIfValid("plan_code", PlanCode);
             xmlWriter.WriteIfCollectionHasAny("subscription_add_ons", AddOns);
-            xmlWriter.WriteStringIfValid("coupon_code", _couponCode);
+            xmlWriter.WriteStringIfValid("coupon_code", _couponCode);
+
+            // TODO change Quantity to int? in next version
+            if (Quantity > 0)
```
bhelx: I'd like to change Quantity to an optional instead of doing this but I'm going to do a follow up for that on the 1.5 branch.

bhelx: We might want to do the same with the other integers actually.

bhelx:
@1TheMuffinMan I added back the WIP because I want to investigate the idea of having a `SubscriptionChange` class the same way as our java client does. Still good for review but I might come back today with a better interface.

1TheMuffinMan:
@bhelx I realize the API supports only accepting the fields that changed, but is there really any harm in modifying the existing logic to send over the full subscription xml? This way you'd be able to keep the same interface and it would be backwards compatible. (I apologize if I already asked this and got a response in a previous conversation)

bhelx:
@1TheMuffinMan That's a good question. The problem boils down to the fact that this is not really a traditional `UPDATE` in the `CRUD` sense. I think the java library expresses this best by having a separate `SubscriptionChange` object that you have to create. Every field that you send expresses an explicit intent to carry over to a new subscription. If you were to send every field over, the API wouldn't be able to distinguish what the programmer wanted to carry over and what he did not.

1TheMuffinMan:
I think that the best approach is to send only what you need to change since the API supports it, so I'm onboard with the argument you're making for the `SubscriptionChange`. 

However, for sake of argument on a low level I still consider changing the subscription as a regular update. If the existing subscription has Qty of 1 and you resend the same value then the new subscription keeps the same value so the API would not need to determine what changed and what didn't.

1TheMuffinMan:
@bhelx Running into a potential design issue that I'm not sure how to go around just yet.

Since the the WriteSubscriptionChangeXml gets executed by the client class later on there isn't a straightforward way to give it dynamic arguments to write within the Subscription class.

The Change method I was working on would take an array of predicates to determine which fields to change: `ChangeSubscription(ChangeTimeFrame.Now, x=> x.Uuid, x=> x.Plan...)`

Any thoughts?

Library/Subscription.cs
```diff
@@ -478,6 +473,17 @@ public bool UpdateNotes(Dictionary<string, string> notes)
             return true;
         }
 
+        /// <summary>
+        /// Generates a fresh subscription object which you can set changes on
+        /// </summary>
+        public Subscription GenerateChange()
```
bhelx: We should remove this now that it's no longer needed.

Library/Subscription.cs
```diff
@@ -422,33 +422,28 @@ public void Preview()
         /// Preview the changes associated with the current subscription
         /// </summary>
         /// <param name="timeframe">ChangeTimeframe.Now (default) or at Renewal</param>
-        public virtual Subscription PreviewChange(ChangeTimeframe timeframe)
+        /// <param name="preds">specify the fields that are changed and should be sent to the server, 
+        /// null sends all fields for backwards compatibility.</param>
+        public virtual Subscription PreviewChange(ChangeTimeframe timeframe, params Expression<Func<Subscription, object>>[] preds)
```
bhelx: Do you think it makes more sense naming the parameter `predicates` instead of `preds`? Longer name but clearer for the person implementing.

bhelx:
@1TheMuffinMan Can we deal with the conflicts? I think it might be easier to do this on a new branch and we can abandon my changes.

aaron-suarez:
Closing this out in favor of #363

PR #231, opened 2017-05-14
Fixing NullReferenceException in List() functions

jakenuts:
Replaces the filter override code in List() functions with a similar code construct that avoids the NullReferenceException caused by calling filter.Equals(null). This error might be specific to newer .NET framework versions but it hobbles much of the functionality in the current packaged release.


bhelx:
Thanks for picking this up @jakenuts . I will review on Monday.

bhelx:
Thanks @jakenuts. I'll get a release out this week.

Issue #233, opened 2017-05-25
Account Balance call to the API is not available in the C# client


argystu:
I was looking for a way to get the account balance in the client.  While the API supports it (v2 - https://dev.recurly.com/docs/lookup-account-balance), the c# client does not.  Am I missing something?  Seems like it should be pretty easy to add.

1TheMuffinMan:
@argystu Thanks for inquiring about this. I do not see it in the library either. I may be able to get this functionality implemented this week.

@bhelx Looks like this has been available since before API V2.5, but didn't make its way into the C# library. Seems like a simple implementation. I was thinking of pausing on subscription change object and completing this tonight. Thoughts?

bhelx:
@1TheMuffinMan that sounds good. This was in api v2.3 I think. It's a pretty straightforward change. I believe in the other libraries I created an `AccountBalance` model for this.

1TheMuffinMan:
@argystu I will work on getting this in tonight for you.

bhelx:
Thanks @1TheMuffinMan

I'll hold off the release for this last feature.

argystu:
@1TheMuffinMan - thank you much!  I really appreciate it.

bhelx:
@1TheMuffinMan You got an ETA on this feature? Would like to get out a release today but I can push to Monday.

1TheMuffinMan:
@argystu Just wanted to let you know that I started working on this last night and hope to have it done by Tuesday.

argystu:
@1TheMuffinMan - Thanks for the heads up.

argystu:
@bhelx - will you be putting out a new release now that this is complete?

bhelx:
@argystu This will go out in 1.5.0: https://github.com/recurly/recurly-client-net/pull/235

I'm aiming for a release tonight. Tomorrow at latest.

bhelx:
The release will be referenced here. Closing this issue now.

PR #234, opened 2017-05-31
Added account balance endpoint

1TheMuffinMan:
references #233


Test/AccountTest.cs
```diff
@@ -146,5 +147,19 @@ public void GetAccountNotes()
 
             notes.Should().BeEmpty();
         }
+
+        [RecurlyFact(TestEnvironment.Type.Integration)]
+        public void Balance()
```
1TheMuffinMan: @bhelx With this test I get back a balance, but the API is not sending back the `PastDue`. How can I modify this test so that I can test the population of that property as well?

bhelx: Well, I think it won't send that field unless it applies. I'm not sure I can think of a way to make an account in a past due state. I'll look into it tomorrow.

bhelx:
Thanks @1TheMuffinMan !

1TheMuffinMan:
@bhelx 
Example documentation:

```
Accounts.Balance("accountCode");

//or

var account = Accounts.Get("accountCode");
account.Balance;
```

PR #230, opened 2017-05-05
Fixed a probable NRE in Plans.List()

CodingGorilla:
Addresses #229

PR #235, opened 2017-06-02
Bump 1.5.0

bhelx:
Going to hold off on this release until I can figure out how to make make this work with removal of `X-Records`

* Added account balance endpoint  #234 (issue #233)
* Added plan and subscription changes  #232
* Fixing NullReferenceException in List() functions  #231
* Fixed a probable NRE in Plans.List() #230
* Remove X-Records header #236 

### Upgrade Notes

This release will upgrade us to API version 2.6. There are two breaking changes:

1. To speed up your listing requests we’re no longer automatically computing the record counts for each requests. For our larger sites this could halve the response time. So in this release, we are removing the `RecurlyList#Capacity` method.
to be cached for you. For more info see [PR #324](https://github.com/recurly/recurly-client-ruby/pull/324).
2. For `POST /v2/subscriptions` Sending `null` for `total_billing_cycles` attribute will now override plan `total_billing_cycles` setting and will make subscription renew forever.
Omitting the attribute will cause the setting to default to the value of plan `total_billing_cycles`.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,14 @@
 Unreleased
 ===============
 
+1.5.0 (stable) / 2017-06-02
+===============
+
+* Added account balance endpoint
+* Added plan and subscription changes
+* Fixing NullReferenceException in List() functions
+* Fixed a probable NRE in Plans.List()
+
 1.4.13 (stable) / 2017-05-04
 ===============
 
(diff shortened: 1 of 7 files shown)


bhelx:
:+1: Ready to go

PR #244, opened 2017-09-20
update postpone timestring for 24-hour formatting

elliottbernstein:
update postpone timestring for 24-hour formatting


bhelx:
:+1: Thanks @elliottbernstein !

Issue #249, opened 2017-10-24
No way to redeem gift cards?


jjwhite:
It would appear that there is currently no way to redeem a gift card using the .NET client?  I noticed that the documentation does not show a C# option (https://dev.recurly.com/docs/redeem-gift-card) and I don't see anything in the GiftCard.cs file that would indicate that you can currently redeem with this library.  Am I missing something or is this a feature that is still pending.

bhelx:
@jjwhite We still find missing features occasionally. I will take a look at this today and if needed, I'll submit a PR.

Also, I should note that if you are attempting to redeem a gift card as part of a signup or a purchase, you should consider looking at the new `/purchases` endpoint: https://dev.recurly.com/docs/create-purchase 

It's an all-in-one endpoint for subscriptions, adjustments, redemptions, etc. If you check the .NET example it includes a gift card redemption.

bhelx:
So, I'm looking at [PR #241](https://github.com/recurly/recurly-client-net/pull/241) for GiftCard support. It does look like we missed that endpoint. I will put this on my schedule for today.

bhelx:
@jjwhite Let me know if #250 looks good to you

bhelx:
Closing with #250

PR #250, opened 2017-10-24
Adds giftcard redeem endpoint

bhelx:
Addresses #249 

```csharp
// first we create the gift card
var delivery = new Delivery(Delivery.DeliveryMethod.Email);
delivery.GifterName = "Benjamin";
delivery.FirstName = "Sally";
delivery.LastName = "DuMonde";
delivery.PersonalMessage = "Happy birthday!";
delivery.EmailAddress = "sally@example.com";

// the first param can be an account code for the gifter (if it exists) or an Account object for a new account
var giftCard = new GiftCard("benjamin-du-monde", delivery, "test_gift_card", 1000, "USD");

giftCard.Create();

Console.WriteLine(giftCard);
Console.WriteLine(giftCard.RedemptionCode);

// Now let's create sally's account
var recipientAccount = new Account("sally-du-monde");

// you don't need to call create here if the account already exists
recipientAccount.Create();

// Now let's redeem the gift card on sally's account
// this will update the giftCard object
giftCard.Redeem(recipientAccount.AccountCode);

Console.WriteLine(giftCard.RedeemedAt);
```

### DIFF
--- a/Library/Account.cs
+++ b/Library/Account.cs
@@ -506,6 +506,17 @@ namespace Recurly
 
             xmlWriter.WriteEndElement(); // End: account
         }
+        /// <summary>
+        /// This serializer is used for redeeming a gift card on
+        /// this account.
+        /// </summary>
+        /// <param name="xmlWriter"></param>
+        internal void WriteGiftCardRedeemXml(XmlTextWriter xmlWriter)
+        {
+            xmlWriter.WriteStartElement("recipient_account"); // Start: recipient_account
+            xmlWriter.WriteElementString("account_code", AccountCode);
+            xmlWriter.WriteEndElement(); // End: recipient_account
+        }
 
         #endregion
 
--- a/Library/GiftCard.cs
+++ b/Library/GiftCard.cs
@@ -150,6 +150,18 @@ namespace Recurly
         }
 
         /// <summary>
+        /// Redeem this gift card on the account
+        /// with the given account code.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


jjwhite:
I still see one problem. If a user is redeeming a gift card they are going to be entering a redemption code so I'd need to be able to create a GiftCard object by passing in the redemption code rather than the delivery information since I wouldn't have access to that at the time of redemption.  So, I think we would need something like this: 

```c#
var giftCard = new GiftCard("my-redemption-code");
giftCard.Redeem(recipientAccount.AccountCode);
```

Here is the python example from the documentation that lines up with what I'm trying to do:
```python
account_code = '3345634'
redemption_code = '22FD95994D02B39D'
gift_card = recurly.GiftCard(redemption_code=redemption_code)
gift_card.redeem(account_code)
```

bhelx:
@jjwhite Thanks for pointing that out. I forgot that that GiftCard initializer was internal. I'll change it to public and it should work. Give me a moment to confirm.

bhelx:
Okay, just pushed that change. I confirmed this code works:

```csharp
var giftCard = new GiftCard("20IX974AEJLGUON8");
giftCard.Redeem("sally-du-monde");
Console.WriteLine(giftCard.RedeemedAt);
```

Does that get you what you need?

jjwhite:
That should work.  I'll let you know if I run into any issues.  Thanks!

bhelx:
No problem. I'll do a bump release after it's merged. If you follow this PR you'll get notified on release.

PR #260, opened 2017-11-21
Fix creation of an empty invoice when InvoiceList is empty

igor-shevchenko:
Right now when the API returns an empty list of invoices, the library parses it as a list with one empty invoice.


bhelx:
Thanks so much!

PR #268, opened 2017-12-11
fix unit amounts exceptions when using percentage addons

danips5150:
When we create a usage addon whose usage type is percentage, the UnitAmountInCents is not required; the mandatory value is the percentage for the addon. In recurly administration console when you select the usage type by percentage, the UnitAmountInCents fields are not shown. 
The problem we saw in "SubscriptionAddOn" and "SubscriptionAddOnList" classes, is that when managing the addons some exceptions are thrown because they are accessing the UnitAmountInCents always, in spite of the fact that the addons have that value as null, because they are defined as percentage (UnitAmountInCents is not needed)

### DIFF
--- a/Library/SubscriptionAddOn.cs
+++ b/Library/SubscriptionAddOn.cs
@@ -1,10 +1,12 @@
-﻿using System.Xml;
+﻿using System;
+using System.Xml;
 
 namespace Recurly
 {
     public class SubscriptionAddOn : RecurlyEntity
     {
         public string AddOnCode { get; set; }
+        public AddOn.Type? AddOnType { get; set; }
         public int UnitAmountInCents { get; set; }
         public int Quantity { get; set; }
         public Adjustment.RevenueSchedule? RevenueScheduleType { get; set; }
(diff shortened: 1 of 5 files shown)


bhelx:
Thanks @danips5150  I will take a look at it this week.

danips5150:
Hi @bhelx !! 

Any news about the pull request? Bad dates I suppose...hehe.

Best regards and merry christmas!!

bhelx:
Thanks for reminding me @danips5150

Library/SubscriptionAddOn.cs
```diff
@@ -14,9 +16,10 @@ public SubscriptionAddOn(XmlTextReader reader)
             ReadXml(reader);
         }
 
-        public SubscriptionAddOn(string addOnCode, int unitAmountInCents, int quantity = 1)
+        public SubscriptionAddOn(string addOnCode, AddOn.Type addOnType, int unitAmountInCents, int quantity = 1)
```
bhelx: I'm not sure this is the way to go here because it breaks the contract. Maybe instead of changing the `SubscriptionAddOn` constructor, we should add another one. Something like this:

```csharp
// keep old constructor
public SubscriptionAddOn(string addOnCode, int unitAmountInCents, int quantity = 1)
{
  AddOnCode = addOnCode;
  UnitAmountInCents = unitAmountInCents;
  Quantity = quantity;
}

// add new constructor
public SubscriptionAddOn(string addOnCode, AddOn.Type addOnType, int unitAmountInCents, int quantity = 1)
{
  AddOnCode = addOnCode;
  AddOnType = addOnType;
  UnitAmountInCents = unitAmountInCents;
  Quantity = quantity;
}
```

What do you think?

danips5150:
@bhelx  requested changes done!! ;-)

danips5150:
Hi @bhelx. Any news about the PR?

bhelx:
@danips5150 Sorry for the delay. Thanks so much for contributing this. When the next release goes out it will be linked here.

PR #271, opened 2018-01-08
Set Errors variable to a default instance of the Errors class.

andrewg187:
A NullReference exception is getting thrown inside the Exception class due a default instance not being set. 
The RecurlyException class is expecting that the passing parameter is always initialized but this is not always the case: 
image

The exception gets triggered when an exception occurs trying parse the XML response here:
image

The NullReference exception is masking other errors occurring in the consumer so this fix will ensure a default instance is always set.

### DIFF
--- a/Library/Errors.cs
+++ b/Library/Errors.cs
@@ -52,7 +52,7 @@ namespace Recurly
             if (response == null)
                 return null;
 
-            Errors errors = null;
+            Errors errors = new Errors();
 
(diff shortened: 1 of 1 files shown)


bhelx:
@andrewg187 thanks for the PR and detailed summary! I will review this today.

andrewg187:
Thank you!

andrewg187:
Thanks for the quick turnaround!

PR #279, opened 2018-03-04
introduce Account Acquisition

pshastov:


pshastov:
It seems **currency** parameter is actually mandatory.

Library/AccountAcquisition.cs
```diff
@@ -0,0 +1,167 @@
...
+        public int? CostInCents { get; set; }
+        public string Currency { get; set; }
+        public AccountAcquisitionChannel? Channel { get; set; }
+        public string SubChannel { get; set; }
+        public string Campaign { get; set; }
+
+        private const string UrlPrefix = "/accounts/";
+        private const string UrlPostfix = "/acquisition";
+
+        public AccountAcquisition(string accountCode) : this()
```
bhelx: Perhaps we want to add Currency to these since it's mandatory?

pshastov: According to the docs, Currency is optional. https://dev.recurly.com/docs/create-account-acquisition
While testing I've run into the issue if Currency not set. Currently, I can't reproduce it.
I've modified CreateAccountAcquisitionWithMandatoryParameters test.

bhelx:
@pshastov Is this working for you and ready to go? I can test and merge if so. Code looks fine.

pshastov:
Yes, it is ready to go

pshastov:
No problem. Thanks for an awesome product

PR #289, opened 2018-04-03
fix typo in Subscription#pause

oktaras:
Hi, when I try to pause a subscription, I get an error from the server that the field 'remaining_pause_cycles' is required. After these changes, the pause is working.

### DIFF
--- a/Library/Subscription.cs
+++ b/Library/Subscription.cs
@@ -569,6 +569,7 @@ namespace Recurly
 
                 DateTime dateVal;
                 Int32 billingCycles;
+                Int32 pauseCycles;
 
                 switch (reader.Name)
                 {
@@ -694,7 +695,7 @@ namespace Recurly
                         break;
 
                     case "remaining_pause_cycles":
-                        if (Int32.TryParse(reader.ReadElementContentAsString(), out int pauseCycles))
+                        if (Int32.TryParse(reader.ReadElementContentAsString(), out pauseCycles))
                             RemainingPauseCycles = pauseCycles;
                         break;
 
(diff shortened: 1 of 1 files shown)


bhelx:
Thanks @oktaras . That code did seem to be incorrect. I'll get this reviewed and publish a bug fix.

Library/Subscription.cs
```diff
@@ -692,6 +693,11 @@ internal override void ReadXml(XmlTextReader reader)
                             RemainingBillingCycles = billingCycles;
                         break;
 
+                    case "remaining_pause_cycles":
+                        if (Int32.TryParse(reader.ReadElementContentAsString(), out int pauseCycles))
```
bhelx: There appears to be an issue here when travis is running the test:
```
Subscription.cs(697,85): error CS1644: Feature `declaration expression' cannot be used because it is not part of the C# 6.0 language specification
```

bhelx: I think it's because `pauseCycles` needs to be declared.

oktaras: @bhelx i can change, but this code works in c# 7.0 and higher https://blogs.msdn.microsoft.com/dotnet/2016/08/24/whats-new-in-csharp-7-0/

bhelx: I understand, but we need it to continue to work in older versions.

oktaras: Ok, I will change.

bhelx:
Thanks so much for taking the time to fix this @oktaras. I'll get the bug fix out today.

oktaras:
Thanks for quick problem resolve.

bhelx:
This was released as `1.11.2`. It should show up in nuget soon: https://github.com/recurly/recurly-client-net/releases/tag/1.11.2

PR #291, opened 2018-04-05
Fix missing `company` field when updating `BillingInfo` instances

### LINKED ISSUE #292: Updating a BillingInfo instance ignores the Company value
The `Company` property is currently being ignored when saving a `BillingInfo` (which appears to result in any existing value being cleared on the server).

See PR #291 for suggested fix.

davidgeary:
Resolves #292

Now includes the `company` field when writing the XML for updating a `BillingInfo` instance.

Also added a `Debug` statement to identify any unhandled fields during reading of XML (noticed that `currency` field isn't currently included but don't think that can be tested unless the environment is configured for multi-currency).

Related tests expanded to include all relevant `BillingInfo` properties.

### DIFF
--- a/Library/BillingInfo.cs
+++ b/Library/BillingInfo.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Diagnostics;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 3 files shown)


bhelx:
Thanks @davidgeary

PR #298, opened 2018-05-14
Fix unit amount in cents for usage add ons

danips5150:
We are getting the following error when we try to create a new subscription with a Usage addon type:

Message: {usagePriceInDecimalUnits is not allowed when PricingType is USAGE. | Field: "subscription.base" Code: "" Symbol: "field_mismatch_not_allowed" Details: ""} 

The problem is that the field unitamountincents was being provided with 0 automatically. With this pull request fix this error.

Regards,


bhelx:
👍

Issue #312, opened 2018-06-26
Subscription.Coupon throws an ArgumentNullException, if _couponCode is null.


jamesbar2:
When calling [`Subscription.Coupon`](https://github.com/recurly/recurly-client-net/blob/master/Library/Subscription.cs#L202), the Recurly client will make a request to fetch the coupon, even if the underlying `_couponCode` is null.  This might not be an issue, except while forming request in [`Coupon.Get(couponCode)`](https://github.com/recurly/recurly-client-net/blob/master/Library/Coupon.cs#L538), `Uri.EscapeDataString(couponCode)` is called without ever checking if the `couponCode` is null or empty.  This results in an `ArgumentNullException` from `Uri`.

To patch this on our end, I'm surrounding `Subscripton.Coupon` with a try/catch(ArgumentNullException)`.  I'll look at creating a PR to fix the issuethough.

bhelx:
@jamesbar2 I think you are correct. It should have a case for when `_couponCode` is null. I'd be happy to review any PR you make for this. If you don't have time to get to it, I can put this in my backlog as well.

jamesbar2:
@bhelx I just forked it and am making the change now.  I assume `Coupon` should just return null.

bhelx:
Yeah, that sounds good.

jamesbar2:
@bhelx Posted a PR for it, #313.  Let me know if you'd like anything changed.  Thanks for your help.

bhelx:
@jamesbar2 we really appreciate you taking this on. I have a release planned for today. I'll see if i can squeeze this feature in.

bhelx:
Closed with #313

jamesbar2:
Thanks @bhelx, happy to help 😄

PR #313, opened 2018-06-26
#312 - Subscription.Coupon throws an ArgumentNullException, if _couponCode is null

jamesbar2:
Addresses https://github.com/recurly/recurly-client-net/issues/312 where Subscription.Coupon would fail if the underlying coupon code was null.

* Updates `Coupons.Get(couponCode)` to return null when the provided couponCode is null or whitespace, preventing a call to Recurly. 
* Updates `Subscription.Coupon` to return null if `_couponCode` is null or whitespace.  However, this does not set the underlying `_coupon` object in the event `_couponCode` is updated elsewhere.
* Adds `TestEnvironment.Type.Unit` for tests, since the tests added do not rely on a web call out.
* Adds tests to `CouponTest` that verifies null is returned for the three scenarios provided, and that the functionality remains consistent.
* Adds tests to `SubscriptionTest` to verify that `Subscription` will receive null, rather than an error when the `_couponCode` is not set.

### DIFF
--- a/Library/Coupon.cs
+++ b/Library/Coupon.cs
@@ -532,6 +532,11 @@ namespace Recurly
         /// <returns></returns>
         public static Coupon Get(string couponCode)
         {
+            if (string.IsNullOrWhiteSpace(couponCode))
+            {
+                return null;
+            }
+
             var coupon = new Coupon();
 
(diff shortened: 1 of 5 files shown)


bhelx:
This will be released in `1.12.2` by the end of the day.

PR #314, opened 2018-06-26
Bump 1.12.2

bhelx:
This release brings us to API version 2.13 but does not have any breaking changes.

* Allow programmer to set gateway code per purchase #303 
* Add link to all_transactions on Invoice #302 
* Subscription Terms #311 
* Subscription.Coupon throws an ArgumentNullException, if _couponCode is null. Now returns null. #313

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,16 @@
 Unreleased
 ===============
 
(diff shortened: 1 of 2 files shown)


bhelx:
👍

PR #319, opened 2018-07-06
Allow more nulls in BillingInfo

bhelx:
This PR allows `CardType`, `Month`, and `Year` to be null. It prevents exceptions when parsing.


bhelx:
Verified the tests are passing locally. There is an issue with Mono on Travis that is causing the failures. Will send a PR to address this.

PR #328, opened 2018-08-24
Allow updating custom fields through /notes endpoint

aaron-suarez:
The changes at lines 332, 338, 347, 352 and 358 remove an invisible character that is unrecognized by OSX. Since it's all in the comments, I figured it's not a problem to remove that character.

### DIFF
--- a/Test/SubscriptionTest.cs
+++ b/Test/SubscriptionTest.cs
@@ -400,12 +400,15 @@ namespace Recurly.Test
             notes.Add("CustomerNotes", "New Customer Notes");
             notes.Add("TermsAndConditions", "New T and C");
             notes.Add("VatReverseChargeNotes", "New VAT Notes");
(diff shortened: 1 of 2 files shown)


aaron-suarez:
Added test

Test/SubscriptionTest.cs
```diff
@@ -400,12 +400,15 @@ public void UpdateNotesSubscription()
             notes.Add("CustomerNotes", "New Customer Notes");
             notes.Add("TermsAndConditions", "New T and C");
             notes.Add("VatReverseChargeNotes", "New VAT Notes");
+            sub.CustomFields = new List<CustomField>(new CustomField[] { new CustomField("food", "taco") });
```
bhelx: I think this can be shortened to:

```csharp
sub.CustomFields.Add(new CustomField("food", "taco"))
```

PR #335, opened 2018-09-11
Add tokenization support for billing info

aaron-suarez:
Adds gateway_code, gateway_token, and card_type in billing info xml.

Playground script:

```C#
// Assumes account "x" exists on this site
// Assumes gateway_code "jhpqd551excn" is the code for vantiv on this site
using System;
using Recurly;

namespace TestRig
{
    class CreateBillingInfo
    {
        {
            var account = Accounts.Get("x");
            var info = new BillingInfo(account);
            info.FirstName = "John";
            info.LastName = "Smith";
            info.Address1 = "123 Main St";
            info.City = "San Fransisco";
            info.State = "CA";
            info.Country = "US";
            info.PostalCode = "90212";
            info.GatewayToken = "4111115851581111";
            info.GatewayCode = "jhpqd551excn";
            info.CardType = BillingInfo.CreditCardType.Visa;
            info.ExpirationMonth = 10;
            info.ExpirationYear = 2020;

            try
            {
                info.Create();
            }
            catch(RecurlyException e)
            {
                Console.WriteLine(e.Errors[0]);
            }
        }
    }
}
```

### DIFF
--- a/Library/BillingInfo.cs
+++ b/Library/BillingInfo.cs
@@ -16,7 +16,7 @@ namespace Recurly
             AmericanExpress,
             Discover,
             JCB,
-            Danokrt,
+            Dankort,
             Maestro,
             Forbrugsforeningen,
             Laser,
@@ -132,6 +132,10 @@ namespace Recurly
         public string VerificationValue { get; set; }
 
         public string TokenId { get; set; }
+
+        public string GatewayToken { get; set; }
+
+        public string GatewayCode { get; set; }
 
         /// <summary>
         /// Timestamp representing the last update of this billing info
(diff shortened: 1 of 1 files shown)


Library/BillingInfo.cs
```diff
@@ -401,6 +405,13 @@ internal override void WriteXml(XmlTextWriter xmlWriter)
                     xmlWriter.WriteElementString("external_hpp_type", ExternalHppType.Value.ToString().EnumNameToTransportCase());
 
                 }
+
+                if (!GatewayCode.IsNullOrEmpty())
+                {
+                    xmlWriter.WriteElementString("gateway_code", GatewayCode);
+                    xmlWriter.WriteElementString("gateway_token", GatewayToken);
+                    xmlWriter.WriteElementString("card_type", CardType.ToString().ToLower());
```
bhelx: Will this produce the correct string for every value? You may need to use `EnumNameToTransportCase` here.

aaron-suarez: Fixed

Library/BillingInfo.cs
```diff
@@ -401,6 +405,15 @@ internal override void WriteXml(XmlTextWriter xmlWriter)
                     xmlWriter.WriteElementString("external_hpp_type", ExternalHppType.Value.ToString().EnumNameToTransportCase());
 
                 }
+
+                if (!GatewayCode.IsNullOrEmpty())
+                {
+                    xmlWriter.WriteElementString("gateway_code", GatewayCode);
+                    xmlWriter.WriteElementString("gateway_token", GatewayToken);
+                    var card = CardType.ToString().EnumNameToTransportCase();
```
bhelx: Can we drop a comment here explaining why this is needed?

aaron-suarez: Yup

PR #337, opened 2018-09-12
Update readme with new config and .net target instructions

bhelx:


README.md
```diff
@@ -33,29 +33,49 @@ For more information about getting started with git, please check out the
...
+Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, pageSize);
 ```
 
+We recommend doing this for compatibility with newer .NET frameworks as well as security. We also recommend you encrypt
+your api key at rest or use a secrets service. *You should not store your api key in plain text on your system.*
+
+#### Configuration Security Note
+
+If you are using the legacy method of configuring with `app.config` or `web.config`, we recommend updating your
+application to the method recommend above.
```
aaron-suarez: I don't think "method recommend above" is proper grammar, plus using "recommend" twice in one sentence is kind of redundant. How about we change this whole sentence to "If you are using the legacy method of configuring with `app.config` or `web.config`, we strongly recommend updating your application to use the above method."

README.md
```diff
@@ -33,29 +33,49 @@ For more information about getting started with git, please check out the
...
+
 ## Client Documentation
 
 The API documentation is available on our [developer docs site](https://dev.recurly.com/docs/getting-started)
 You can find .NET examples there but we have some examples in [examples.md](./examples.md).
 
+## .NET core and standard support
+
+Although we do not officially support .NET core or standard at this moment, it is possible to use this library
+in those applications. We recommend vendoring the code, building the dlls, and linking directly to them from
```
aaron-suarez: I don't think "recommend" is the right word here. How about "It is possible to vendor the code..."

PR #345, opened 2018-10-10
Change formatting of Usage dates in WriteXML to match Recurly format

### LINKED ISSUE #338: Datetime Parsing not utc
https://github.com/recurly/recurly-client-net/blob/60e750f075a3c5ce2ca5f4574c593edfaa83ba57/Library/Usage.cs#L140

Currently  I am trying the following code to update the amount for usage based addons 
```
 var usage = Recurly.Usages.List(subUuid, addonCode).OrEmpty().FirstOrDefault();
  if(usage!=null)
  {
         
            usage.Amount = quantity;
            usage.Update();
  }

```
I think the issue is that Datetime.Parse used to get recording timestamp and usage time timestamp doesn't return date in UTC . However when in the WriteXML it is just converted using DatetimeToString("s").  So if I am passing a local date time , which is something I got when I retrieved the usage , it throws an error. I think API expects it in UTC and the current time in AEST when read as UTC will be in future and it throws an error.

This forces me to write the above code as  so that timestamps are ignored
```
{  
         
           usage.Amount = quantity;
           usage.UsageTimestamp = null;
           usage.RecordingTimestamp = null;
            usage.Update();
  }
```

Is there a better fix for this so that I don't have to set the properties which I don't want to update

For eg : 

> DateTime.Parse("2018-09-13T00:43:14Z").ToString("s")  

will give the result 

> 2018-09-13T10:43:14 

instead of 

> 2018-09-13T00:43:14Z

aaron-suarez:
Fixes #338 

Playground script for testing. Be sure to change the sub UUID and Addon code to whatever your sandbox site uses. Tested with no updates to the recording time, using DateTimeKind.Utc as well as Local and Unspecified.

```C#
using System;
using Recurly;
using System.Linq;

namespace TestRig
{
    class UpdateUsage 
    {
        {
            var subscriptionUuid = "47c92b1ea14c0a67ed60fd4f01bded5f";
            var subscriptionAddOnCode= "usage_based";
            var quantity = 2;
            var recTime = new DateTime(2018, 10, 10, 2, 30, 20, DateTimeKind.Local);

            var usage = Recurly.Usages.List(subscriptionUuid, subscriptionAddOnCode).FirstOrDefault();
            if(usage!=null)
            {
                    usage.Amount = quantity;
                    usage.RecordingTimestamp = recTime;
                    usage.Update();
            }
        }
    }
}
```

### DIFF
--- a/Library/Usage.cs
+++ b/Library/Usage.cs
@@ -189,13 +189,13 @@ namespace Recurly
             xmlWriter.WriteElementString("merchant_tag", MerchantTag);
             
             if (RecordingTimestamp.HasValue)
-                xmlWriter.WriteElementString("recording_timestamp", RecordingTimestamp.Value.ToString("s"));
+                xmlWriter.WriteElementString("recording_timestamp", RecordingTimestamp.Value.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ"));
 
             if (UsageTimestamp.HasValue)
-                xmlWriter.WriteElementString("usage_timestamp", UsageTimestamp.Value.ToString("s"));
+                xmlWriter.WriteElementString("usage_timestamp", UsageTimestamp.Value.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ"));
 
             if (BilledAt.HasValue)
-                xmlWriter.WriteElementString("billed_at", BilledAt.Value.ToString("s"));
+                xmlWriter.WriteElementString("billed_at", BilledAt.Value.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ"));
 
             xmlWriter.WriteEndElement();
         }


Library/Usage.cs
```diff
@@ -189,13 +189,13 @@ internal void WriteXml(XmlTextWriter xmlWriter, bool update)
             xmlWriter.WriteElementString("merchant_tag", MerchantTag);
             
             if (RecordingTimestamp.HasValue)
-                xmlWriter.WriteElementString("recording_timestamp", RecordingTimestamp.Value.ToString("s"));
+                xmlWriter.WriteElementString("recording_timestamp", RecordingTimestamp.Value.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ"));
```
bhelx: Was there not a standard ISO8601 format code? why spell it out here?

aaron-suarez: The [format code for "o"](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#the-round-trip-o-o-format-specifier) is the one called out in the Microsoft docs for this standard. If you use that, it looks like this `2018-10-10T07:30:20.0000000Z`

The only real difference is all of those trailing `0`s which I suppose isn't an issue. I wanted it to match the value coming back from Recurly exactly, but I can change the format to "o" if you think that makes more sense for this library.

Issue #146, opened 2016-03-29
Remove examples.md?


bhelx:
I think maybe we should remove the `examples.md` file. I can't say for sure whether they all work It hasn't been updated. We should direct people to find examples at the docs site where we can keep things up to date. It's either that or we go through every example and check that they still work then keep it up in two places.

https://github.com/recurly/recurly-client-net/blob/master/examples.md

Opinions?

\cc @csmb @MuffinMan10110

csmb:
Good catch! I think we should take a look at what it offers, and perhaps verify/update example code on `dev.recurly.com` with what's in `examples.md`.

MuffinMan10110:
The problem with keeping examples on github is that there's always two places to update. Directing people to the docs on the Recurly site sounds like a better option to me.

bhelx:
@MuffinMan10110 I agree. Too hard to keep up with

@csmb I'm going to do that client lib example code audit soon!

1TheMuffinMan:
@bhelx Can I remove examples.md or should I replace it with links to the documention:

API docs
https://dev.recurly.com/

.NET client docs
https://dev.recurly.com/page/net

PR #354, opened 2018-11-18
Updated examples.md

devigo:


aaron-suarez:
@devigo We've made the decision to remove the examples.md file per #146. Before we can remove it, we need to ensure that dev.recurly.com is fully up to date (which I plan on doing in the near future).

For that reason, I'm going to go ahead and close this out. Thanks for your contribution.

PR #356, opened 2018-11-21
Add ExemptionCertificate attribute to Account

aaron-suarez:
Adds exemption_certificate attribute to the account object.

Script:
```c#
using System;
using Recurly;

namespace TestRig
{
    class CreateAccount
    {
        {
            var account = new Account("new-account");
            account.TaxExempt = true;
            account.ExemptionCertificate = "Some Certificate";
            account.Create();
        }
    }
}
```

### DIFF
--- a/Test/AccountTest.cs
+++ b/Test/AccountTest.cs
@@ -33,6 +33,7 @@ namespace Recurly.Test
                 AcceptLanguage = "en",
                 VatNumber = "my-vat-number",
                 TaxExempt = true,
+                ExemptionCertificate = "Some Certificate",
                 EntityUseCode = "I",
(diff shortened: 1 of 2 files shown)


Test/AccountTest.cs
```diff
@@ -53,6 +54,7 @@ public void CreateAccountWithParameters()
             acct.CcEmails.Should().Be("cc1@test.com,cc2@test.com");
             Assert.Equal("my-vat-number", acct.VatNumber);
             Assert.True(acct.TaxExempt.Value);
+            Assert.Equal(acct.ExemptionCertificate, "Some Certificate");
```
aaron-suarez: This test will fail until this feature makes it into production

PR #358, opened 2018-12-03
Add custom fields to WriteChangeSubscriptionXml

aaron-suarez:
There are two different `WriteXml` type functions on the Subscription object. The one for the `/notes` endpoint includes custom fields, but the one for the regular `/subscriptions` endpoint does not. This PR fixes that.

Script:
```C#
    // Assumes "47852d1aefe69b0e9c69504d93ab46f9" is a valid subscription on the account
    // Assumes "device_id" is a configured custom field on the account
    class UpdateSubscription
    {
        {
            var sub = Subscriptions.Get("47852d1aefe69b0e9c69504d93ab46f9");
            sub.CustomFields.Add(new CustomField("device_id", "KIWTL-WER-ZXMRD"));
            sub.ChangeSubscription();
        }
    }
}

### DIFF
--- a/Library/Subscription.cs
+++ b/Library/Subscription.cs
@@ -1062,6 +1062,8 @@ namespace Recurly
 
             if (RenewalBillingCycles.HasValue)
                 xmlWriter.WriteElementString("renewal_billing_cycles", RenewalBillingCycles.Value.AsString());
+            
+            xmlWriter.WriteIfCollectionHasAny("custom_fields", CustomFields);
 
             xmlWriter.WriteEndElement(); // End: subscription
         }

### CODE REVIEW

**Comment by bhelx** [comment]
Can you double check that it allows you to clear a custom field (or all of them) ?

**Comment by aaron-suarez** [comment]
It does not clear a custom field (or all of them). I'm actually not sure what the XML needs to look like in order to clear it out. I've tried some things with Postman (empty Custom Fields array, empty Name and Value in a custom field, etc) and nothing that I try actually seems to clear out any custom fields.

If you can point me to some docs about how to remove a custom field on a subscription via the API (or just describe what the request should look like) I can update the PR to ensure that it allows removal.

**Comment by aaron-suarez** [comment]
Ah, ok it turns out that you have to set the name but set the value to null. This might actually work as is, I'll try it out

**Comment by aaron-suarez** [comment]
Yes, it works to remove custom fields as well. The following removes all custom fields from a subscription:
```C#
var sub = Subscriptions.Get("47852d1aefe69b0e9c69504d93ab46f9");
foreach (var field in sub.CustomFields)
{
    field.Value = null;
}
sub.ChangeSubscription();
```

If you wanted to change one in particular, I'm not sure the best way in C# to grab it without having to loop over the full list. So I'm not saying this is the most efficient way to do the task, but it does indeed work:
```C#
var sub = Subscriptions.Get("47852d1aefe69b0e9c69504d93ab46f9");
foreach (var field in sub.CustomFields)
{
    if (field.Name == "device_id")
        field.Value = null;
}
sub.ChangeSubscription();
```

bhelx:
Right, the way custom fields was designed was to pass in a null value to clear it. This looks good.

PR #359, opened 2018-12-03
Add EnterOfflinePayment method to Invoice

aaron-suarez:
This adds a method to handle this API endpoint https://dev.recurly.com/docs/enter-an-offline-payment-for-a-manual-invoice

The test pretty much shows example usage, but here's a script just in case:
```C#
using System;
using Recurly;

namespace TestRig
{
    class EnterOfflinePayment
    {
        {
            var account = new Account(System.Guid.NewGuid().ToString());
            account.Create();

            var adjustment = account.NewAdjustment("USD", 5000, "Test Charge");
            adjustment.Create();

            var newInvoice = new Invoice();
            newInvoice.CollectionMethod = Invoice.Collection.Manual;

            var invoice = account.InvoicePendingCharges(newInvoice).ChargeInvoice;

            var offlineTransaction = new Transaction(account, 5000, "");
            offlineTransaction.PaymentMethod = "credit_card";
            offlineTransaction.CollectedAt = new DateTime(2018, 12, 3, 2, 30, 20, DateTimeKind.Local);
            offlineTransaction.Description = "Paid the test charge. Gooood customer *pats customer head*";

            try
            {
                var recordedTransaction = invoice.EnterOfflinePayment(offlineTransaction);
                Console.WriteLine(recordedTransaction.Status);
            }
            catch(ValidationException e)
            {
                Console.WriteLine(e);
                foreach (var err in e.Errors) {
                  Console.WriteLine(err);
                }
            }
        }
    }
}
```

### DIFF
--- a/Library/Invoice.cs
+++ b/Library/Invoice.cs
@@ -290,12 +290,12 @@ namespace Recurly
             var refunds = new RefundList(adjustments, options);
             var invoice = new Invoice();
 
-            var response = Client.Instance.PerformRequest(Client.HttpRequestMethod.Post,
+            var statusCode = Client.Instance.PerformRequest(Client.HttpRequestMethod.Post,
                 memberUrl() + "/refund",
                 refunds.WriteXml,
                 invoice.ReadXml);
 
-            if (HttpStatusCode.Created == response || HttpStatusCode.OK == response)
+            if (HttpStatusCode.Created == statusCode || HttpStatusCode.OK == statusCode)
                 return invoice;
             else
                 return null;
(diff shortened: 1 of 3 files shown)


Library/Invoice.cs
```diff
@@ -395,6 +395,26 @@ public Invoice RefundAmount(int amountInCents, RefundMethod method = RefundMetho
...
+        /// <summary>
+        /// Enter an offline payment for a manual invoice
+        /// </summary>
+        /// <param name="transaction">The transaction to be entered.</param>
+        /// <returns>new Transaction object</returns>
+        public Transaction EnterOfflinePayment(Transaction transaction)
+        {
+            var successfulTransaction = new Transaction();
+
+            var response = Client.Instance.PerformRequest(Client.HttpRequestMethod.Post,
```
bhelx: Kind of a nit, but might make more sense to call this `statusCode`. Makes line 412 a little confusing.

aaron-suarez: I copied and pasted from this
https://github.com/recurly/recurly-client-net/blob/38c67ce96637c083aa1c5edd49b2a1ffbbd777c2/Library/Invoice.cs#L293
Do you want me to go ahead and change it to `statusCode` everywhere in the file? I'd rather it be consistent if I change mine.

bhelx: Yeah, I think it makes more sense if it's easy to change.

aaron-suarez: Fixed

PR #363, opened 2018-12-10
Add SubscriptionChange class

### LINKED ISSUE #157: BLOCKER wrong request are send on change subscription

aaron-suarez:
Fixes #157 

The pattern here is different from the patterns of this library, but I believe it will be an improvement because it makes all the changes you want to make explicit, rather than implicit. 

The developer must make a SubscriptionChange object, then any value that ought to be changed should be explicitly set on that object. This object can then be passed into the new static methods `PreviewChange()` and `ChangeSubscription()`, which will return the changed subscription in a new Subscription object.

Script:
```c#
using System;
using Recurly;

namespace TestRig
{
    class PreviewSubscriptionChange
    {
        {
            var change = new SubscriptionChange()
            {
                PlanCode = "gold"
            };
            var newSub = Subscription.PreviewChange("48e99881b3726425bf49e64b45af7fe4", change);
        }
    }
}
```
Another Script:
```c#
using System;
using System.Linq;
using Recurly;
using System.Collections.Generic;

namespace TestRig
{
    class UpdateSubscription
    {
        {
            var sub = Subscriptions.Get("47852d1aefe69b0e9c69504d93ab46f9");
            var subChange = new SubscriptionChange()
            {
                CustomFields = sub.CustomFields
            };

            foreach (var field in subChange.CustomFields)
            {
                if (field.Name == "internal-code")
                    field.Value = null;
            }
            Subscription.ChangeSubscription(sub.Uuid, subChange);
        }
    }
}
```

### DIFF
--- a/Library/Subscription.cs
+++ b/Library/Subscription.cs
@@ -30,12 +30,6 @@ namespace Recurly
             Paused   = 1024
         }
 
-        public enum ChangeTimeframe : short
-        {
-            Now,
-            Renewal
-        }
-
         public enum RefundType : short
         {
             Full,
@@ -423,29 +417,6 @@ namespace Recurly
                 ReadXml);
         }
 
-        /// <summary>
-        /// Request that an update to a subscription take place
-        /// </summary>
-        /// <param name="timeframe">when the update should occur: now (default) or at renewal</param>
-        public void ChangeSubscription(ChangeTimeframe timeframe)
-        {
-            Client.WriteXmlDelegate writeXmlDelegate;
-
-            if (ChangeTimeframe.Renewal == timeframe)
-                writeXmlDelegate = WriteChangeSubscriptionAtRenewalXml;
-            else
-                writeXmlDelegate = WriteChangeSubscriptionNowXml;
-
-            Client.Instance.PerformRequest(Client.HttpRequestMethod.Put,
-                UrlPrefix + Uri.EscapeDataString(Uuid),
-                writeXmlDelegate,
-                ReadXml);
-        }
-
-        public void ChangeSubscription()
-        {
-            ChangeSubscription(ChangeTimeframe.Now);
-        }
 
         /// <summary>
         /// Cancel an active subscription.  The subscription will not renew, but will continue to be active
@@ -482,8 +453,7 @@ namespace Recurly
         /// <summary>
         /// Transforms this object into a preview Subscription applied to the account.
         /// </summary>
-        /// <param name="timeframe">ChangeTimeframe.Now (default) or at Renewal</param>
-        public void Preview(ChangeTimeframe timeframe)
+        public void Preview()
         {
             if (_saved)
             {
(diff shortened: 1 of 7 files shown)


Library/SubscriptionChange.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// for data analysis or a real trial. This value can only be changed on a trial subscription
+        /// and is persisted for its lifetime.
+        /// </summary>
+        public bool? ImportedTrial { get; set; }
+
+        /// <summary>
+        /// Creates a new subscription change object
+        /// </summary>
+        /// <param name="uuid">The uuid of the subscription to be changed</param>
+        public SubscriptionChange(String uuid)
```
bhelx: I like the idea of making it required, but I think it might make more sense to make it a parameter on the method.

```csharp
var change = new SubscriptionChange() {
    PlanCode = "gold"
};
var uuid = "48e99881b3726425bf49e64b45af7fe4";
var newSub = Subscription.PreviewChange(uuid, change);
```

I feel like I could go either way on it. What do you think?

bhelx: I think the reason my gut is telling me to go this way is that it's a URL parameter.

aaron-suarez: I think that you're right, but perhaps for a different reason. The idea of the `SubscriptionChange` class is that each instance of the class would represent the set of changes that need to be sent to the server.

The UUID isn't something that is going to change, it's just something necessary for the endpoint. For that reason, I feel like `Uuid` shouldn't even be a property of the SubscriptionChange class. I think passing it into the method would make a lot of sense

bhelx: Yeah, let's do that.

Library/Subscription.cs
```diff
@@ -587,6 +519,50 @@ public RecurlyList<CouponRedemption> GetRedemptions()
...
+        /// <summary>
+        /// Request that an update to a subscription take place
+        /// </summary>
+        /// <param name="uuid">The uuid of the subscription to be changed</param>
+        /// <param name="change">A subscription change object listing what to change about the subscription</param>
+        public static Subscription ChangeSubscription(String uuid, SubscriptionChange change)
+        {
+            if (uuid.IsNullOrEmpty())
+            {
+                throw new Recurly.RecurlyException("Must have an existing subscription to make changes.");
```
bhelx: I don't know if this is the right exception here. I think something like `ArgumentException` makes more sense. We want to throw an error that the calling code would never catch. If this is null or empty the program or request should explode. The message should say something about what the problem is (that the uuid parameter is null or empty).

Library/SubscriptionChange.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                xmlWriter.WriteElementString("collection_method", "manual");
+                xmlWriter.WriteElementString("net_terms", NetTerms.Value.AsString());
+                xmlWriter.WriteElementString("po_number", PoNumber);
+            }
+            else if (CollectionMethod.Like("automatic"))
+                xmlWriter.WriteElementString("collection_method", "automatic");
+
+            if (ImportedTrial.HasValue)
+            {
```
bhelx: Minor nit, but we should try to apply curly braces consistently. Easiest thing to do would probably be to remove them here.

Library/SubscriptionChange.cs
```diff
@@ -0,0 +1,153 @@
...
+            if (UnitAmountInCents.HasValue)
+                xmlWriter.WriteElementString("unit_amount_in_cents", UnitAmountInCents.Value.AsString());
+
+            if (CollectionMethod.Like("manual"))
+            {
+                xmlWriter.WriteElementString("collection_method", "manual");
+                xmlWriter.WriteElementString("net_terms", NetTerms.Value.AsString());
+                xmlWriter.WriteElementString("po_number", PoNumber);
+            }
+            else if (CollectionMethod.Like("automatic"))
```
bhelx: Another nit, it's kind of awkward to apply braces to one part of the conditional but not the other. Even though it's one line I think it makes more sense to apply braces here.

Library/Subscription.cs
aaron-suarez: Good catch. It made more sense as a `RecurlyException` when the uuid was a class property. Now, `ArgumentException` makes a lot more sense.

Library/SubscriptionChange.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                xmlWriter.WriteElementString("collection_method", "automatic");
+            }
+
+            if (ImportedTrial.HasValue)
+                xmlWriter.WriteElementString("imported_trial", ImportedTrial.Value.ToString().ToLower());
```
sebgod: Shouldn't this be: `.ToLowerInvariant()`?

aaron-suarez: That was a carry-over from Subscription.cs, and was added in https://github.com/recurly/recurly-client-net/commit/2c8755a3f2298b04ee67ed998188e7525f781d56. Has it caused any problems for you?

It probably "should" be `.ToLowerInvariant()` but it's always a boolean so I wouldn't expect there to be any localization issues.

Feel free to open a PR if you think it can be improved, and the current maintainers will review it for you.

PR #372, opened 2019-01-31
Transaction error

PashaSk:
1. added TransactionError property of Transaction
2. case for "transaction_error" node inside parsing cycle
3. TransactionErrorEnum created according to reference from https://dev.recurly.com/page/transaction-errors for using in predicates, e.g.:
`if (transaction.Error.ErrorCodeEnum == TransactionErrorCodeEnum.fraud_risk_check) {
...
}`


Library/TransactionErrorEnum.cs
```diff
@@ -0,0 +1,95 @@
+﻿namespace Recurly
+{
+    /// <summary>
+    /// Transaction Error Codes Enum
+    /// </summary>
+    public enum TransactionErrorCodeEnum
```
aaron-suarez: Does this enum have to be its own separate file? Do you think it might make more sense to make it a part of the `TransactionError` class?

We have several enums embedded within classes throughout this library, so that would keep with the same pattern. E.g.,
https://github.com/recurly/recurly-client-net/blob/7bb333008706c0bb2d529914fe22312207262c0a/Library/Account.cs#L19-L24

aaron-suarez: You're absolutely right, it is rather big. And as a matter of fact, I think it would be difficult to maintain going forward.

`TransactionError` already contains the error code as a string:
https://github.com/recurly/recurly-client-net/blob/36646ccdd6f2ddde5aea98e6876fc56a9d41cbdc/Library/TransactionError.cs#L16

I think the original issue you mentioned is solved by just the changes in `Transaction.cs` (namely, adding a property called `Error` and setting it via the `transaction_error` in the XML from the API).

Users can still check the value in predicates via `if (transaction.Error.ErrorCode == "fraud_risk_check") { ... }`

I know that an Enum would be more performant, but I can't guarantee that these possible values won't change and it seems like it could be a source of errors in the future.

Besides the performance difference between comparing an Enum vs comparing strings, is there any other reason why `ErrorCodeEnum` makes more sense than just having a string `ErrorCode`?

PashaSk: It always good to have all errors codes in one place and put the same reference in predicates. Every time I write a comparison with some hard-coded string, my intuition advises me to make a shared source for all values. Anyway, we could have these references inside our own project.

aaron-suarez: Upon further reflection, I don't think it takes away anything to have these values inside the library. All unexpected cases should be handled by `not_recognized` and in that event, users can go ahead and check the value of transaction.Error.ErrorCode to identify what the unexpected value was. I think I'm good with this as is

PR #376, opened 2019-02-07
Fix issue when usage_percentage is null

### LINKED ISSUE #375: Release 1.15.0 GetSubscriptions() Throws Error When There's A Usage Based Add-On of Type "Price"
**Release 1.15.0** introduced a bug that only occurs when calling **Recurly.GetSubscriptions()** where 1 or more of the returned subscriptions has a usage based add-on attached to it.

The bug is in the **SubscriptionAddOn** class in the **ReadXml** function (see below)  and the offending section is that final case statement. You see, when the usage_type = “price”, the XML returned for usage_percentage is `<usage_percentage nil="nil"/>`. Since this cannot be converted to a float, an error is thrown.

```
internal override void ReadXml(XmlTextReader reader)
        {
            while (reader.Read())
            {
                if (reader.Name == "subscription_add_on" && reader.NodeType == XmlNodeType.EndElement)
                    break;

                if (reader.NodeType != XmlNodeType.Element) continue;

                switch (reader.Name)
                {
                    case "add_on_code":
                        AddOnCode = reader.ReadElementContentAsString();
                        break;

                    case "add_on_type":
                        AddOnType = reader.ReadElementContentAsString().ParseAsEnum<AddOn.Type>();
                        break;

                    case "quantity":
                        Quantity = reader.ReadElementContentAsInt();
                        break;

                    case "unit_amount_in_cents":
                        int unitAmountInCents;
                        if (Int32.TryParse(reader.ReadElementContentAsString(), out unitAmountInCents))
                            UnitAmountInCents = unitAmountInCents;
                        break;

                    case "revenue_schedule_type":
                        var revenueScheduleType = reader.ReadElementContentAsString();
                        if (!revenueScheduleType.IsNullOrEmpty())
                            RevenueScheduleType = revenueScheduleType.ParseAsEnum<Adjustment.RevenueSchedule>();
                        break;

                    case "usage_percentage":
                        UsagePercentage = reader.ReadElementContentAsFloat();
                        break;
                }
            }
        }
```

aaron-suarez:
Fixes #375 

This change reads the tag for `usage_percentage` as a string and then converts it to a float with Invariant culture format so it will be represented the same way regardless of locale settings.

### DIFF
--- a/Library/SubscriptionAddOn.cs
+++ b/Library/SubscriptionAddOn.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Xml;
+using System.Globalization;
 
 namespace Recurly
 {
@@ -70,7 +71,10 @@ namespace Recurly
                         break;
 
                     case "usage_percentage":
-                        UsagePercentage = reader.ReadElementContentAsFloat();
+                        var usagePercentage = reader.ReadElementContentAsString();
+                        if (Single.TryParse(usagePercentage, NumberStyles.Number,
+                                CultureInfo.InvariantCulture.NumberFormat, out float percentage))
+                            UsagePercentage = percentage;
                         break;
                 }
             }


bhelx:
👍

PR #381, opened 2019-03-08
Fix Subscription Change issues regarding AddOns

### LINKED ISSUE #379: Subscription change issue
There are some dangerous issues with the way the _SubscriptionChange_ object works.

If you take this code, for example: 

                var change = new SubscriptionChange 
                { 
                    TimeFrame = SubscriptionChange.ChangeTimeframe.Now,
                    NetTerms = 30,
                    PoNumber = "12345"
                };

                ChangeSubscription(subscriptionUuid, change);

What you would expect: 

- Net terms and PO number are updated to new values.

What actually happens is:

- **All add-ons are removed from the subscription!**
The addon-list defaults to an empty list, instead of _null_. You can't actually _null_ it, because the property getter (which is called when writing the XML) will just recreate an empty list.
- Net terms and PO number are **not** updated. 
You have to also set the _CollectionMethod_ property, even if it didn't change, because of a check in the _WriteChangeSubscriptionXml_ method.

aaron-suarez:
Fixes #379

There were some issues with how SubscriptionAddOns work in this library, and this attempts to fix those issues.

Example script:
```C#
            // Assumes a subscription with uuid "4ac9897d06e86fddcbb62c4a02b54753" 
            // exists on the account
            var subscriptionUuid = "4ac9897d06e86fddcbb62c4a02b54753";

            var change = new SubscriptionChange 
            { 
                TimeFrame = SubscriptionChange.ChangeTimeframe.Now,
                NetTerms = 30,
                PoNumber = "12345",
                AddOns = new SubscriptionAddOnList()
            };

            // Assumes an add on with code "ipaddresses" exists on the account
            var addOnCode = "ipaddresses";
            var quantity = 1;
            var amountInCents = 200;
            change.AddOns.Add(addOnCode, quantity, amountInCents);

            Subscription.ChangeSubscription(subscriptionUuid, change);
```

### DIFF
--- a/Library/List/SubscriptionAddOnList.cs
+++ b/Library/List/SubscriptionAddOnList.cs
@@ -6,6 +6,8 @@ namespace Recurly
     {
         private Subscription _subscription;
 
+        public SubscriptionAddOnList() {}
+
         public SubscriptionAddOnList(Subscription subscription)
         {
             _subscription = subscription;
@@ -64,6 +66,14 @@ namespace Recurly
         public void Add(AddOn planAddOn, int quantity = 1)
         {
             int amount = 0;
+
+            if (_subscription == null)
+            {
+                throw new ValidationException(
+                    "SubscriptionAddOnList must be initialized with a Subscription in order to use this method. Try Add(AddOn planAddOn, int quantity, int unitAmountInCents) instead."
+                    , new Errors());
+            }
+
             if (planAddOn.UnitAmountInCents.Count > 0 && !planAddOn.UnitAmountInCents.TryGetValue(_subscription.Currency, out amount))
             {
                 throw new ValidationException(
@@ -100,6 +110,12 @@ namespace Recurly
         // sub.AddOns.Add(code); 1, unitInCents=this.Plan.UnitAmountInCents[this.Currency]
         public void Add(string planAddOnCode, int quantity=1)
         {
+            if (_subscription == null)
+            {
+                throw new ValidationException(
+                    "SubscriptionAddOnList must be initialized with a Subscription in order to use this method. Try Add(string planAddOnCode, int quantity, int unitAmountInCents) instead."
+                    , new Errors());
+            }
             var unitAmount = _subscription.Plan.AddOns.Find(ao => ao.AddOnCode == planAddOnCode).UnitAmountInCents[_subscription.Currency];
             var sub = new SubscriptionAddOn(planAddOnCode, unitAmount, quantity);
             base.Add(sub);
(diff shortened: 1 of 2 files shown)


Library/List/SubscriptionAddOnList.cs
```diff
@@ -6,6 +6,8 @@ public class SubscriptionAddOnList : RecurlyList<SubscriptionAddOn>
     {
         private Subscription _subscription;
 
+        public SubscriptionAddOnList() {}
```
aaron-suarez: It seems pointless to require a subscription just to check the currency. It made sense with the old pattern, but now with the SubscriptionChange object, you don't always have a Subscription object, and you shouldn't need one. Besides, client side validation is bad.

Library/List/SubscriptionAddOnList.cs
```diff
@@ -64,7 +66,7 @@ internal override void ReadXml(XmlTextReader reader)
         public void Add(AddOn planAddOn, int quantity = 1)
         {
             int amount = 0;
-            if (planAddOn.UnitAmountInCents.Count > 0 && !planAddOn.UnitAmountInCents.TryGetValue(_subscription.Currency, out amount))
+            if (_subscription != null && planAddOn.UnitAmountInCents.Count > 0 && !planAddOn.UnitAmountInCents.TryGetValue(_subscription.Currency, out amount))
```
aaron-suarez: Again, client side validation is bad.

Library/List/SubscriptionAddOnList.cs
```diff
@@ -100,13 +102,25 @@ public void Add(AddOn planAddOn, int quantity, int unitAmountInCents)
         // sub.AddOns.Add(code); 1, unitInCents=this.Plan.UnitAmountInCents[this.Currency]
         public void Add(string planAddOnCode, int quantity=1)
         {
+            if (_subscription == null)
+            {
+                throw new ValidationException(
```
aaron-suarez: If you do want to use the Plan AddOn object instead of just specifying planAddOnCode, quantity, and unitAmountInCents, you need to have a subscription or else you'll run into errors trying to determine the currency.

Library/SubscriptionChange.cs
```diff
@@ -110,14 +108,17 @@ internal void WriteChangeSubscriptionXml(XmlTextWriter xmlWriter)
...
             }
             else if (CollectionMethod.Like("automatic"))
             {
                 xmlWriter.WriteElementString("collection_method", "automatic");
             }
 
+            if (NetTerms.HasValue)
+                xmlWriter.WriteElementString("net_terms", NetTerms.Value.AsString());
+
+            xmlWriter.WriteStringIfValid("po_number", PoNumber);
```
aaron-suarez: Even though the API requires a manual invoice for these values, the client shouldn't be checking to see if the CollectionMethod is manual.

Library/List/SubscriptionAddOnList.cs
aaron-suarez: Since this method takes in a plan add on, I'm probably going to change it to throw a similar error as the other Add methods that take a planAddOn

Library/List/SubscriptionAddOnList.cs
aaron-suarez: I just realized this particular one doesn't actually take in an AddOn object, my above comment was inaccurate. But still, it can't just know what currency to use, so you need a `_subscription` for this to work

Library/SubscriptionChange.cs
```diff
@@ -110,14 +108,17 @@ internal void WriteChangeSubscriptionXml(XmlTextWriter xmlWriter)
             if (CollectionMethod.Like("manual"))
```
bhelx: If we don't need this conditional anymore, i think we can just render the string directly. Something like this:

`xmlWriter.WriteElementString("collection_method", CollectionMethod);`

PR #385, opened 2019-03-19
Fix coupon parsing bug

bhelx:
Two changes here:

1. Remove Coupon#Plans. The API is returning both strings and Lists for this property. This won't be fixed until 2018-10-04.
2. No longer display warnings for serialization errors, instead we should throw an exception.

### DIFF
--- a/Recurly.Tests/JsonSerializerTest.cs
+++ b/Recurly.Tests/JsonSerializerTest.cs
@@ -51,16 +51,13 @@ namespace Recurly.UnitTests
         public void DeserializeWithWrongType()
         {
             var json = "{\"my_string\":\"benjamin\",\"my_int\":\"49urj\"}";
-            var resource = _jsonSerializer.Deserialize<MyResource>(MockResourceResponse(json));
-            // It should ignore the field that was wrong type but still parse other properties
-            Assert.Null(resource.MyInt);
-            Assert.Equal("benjamin", resource.MyString);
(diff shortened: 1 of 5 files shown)


Recurly.Tests/JsonSerializerTest.cs
```diff
@@ -51,16 +51,13 @@ public void Deserialize()
...
-            Assert.Null(resource.MyInt);
-            Assert.Equal("benjamin", resource.MyString);
+            Assert.Throws<JsonReaderException>(() => _jsonSerializer.Deserialize<MyResource>(MockResourceResponse(json)));
         }
 
         [Fact]
         public void DeserializeWithNewUnrecognizedKey()
         {
-            var json = "{\"my_string\":\"benjamin\",\"unregonized\":\"unknown\"}";
+            var json = "{\"my_string\":\"benjamin\",\"unrecognized\":\"unknown\"}";
```
aaron-suarez: Nice catch

Recurly/Resources/Coupon.cs
```diff
@@ -80,10 +80,6 @@ public class Coupon : Resource {
     [JsonProperty("name")]
     public string Name { get; set; }
   
-    /// <value>Plans</value>
-    [JsonProperty("plans")]
-    public List<PlanMini> Plans { get; set; }
```
aaron-suarez: Are we going to add this back in when 2018-10-04 is GA?

aaron-suarez: Nevermind, I just realized this is a generated file and I see exactly how it will be.

PR #390, opened 2019-04-09
Allow configurable api url

bhelx:
This allows reconfiguring the base api url for testing purposes. I tried to make it as unattractive and awkward as possible.

### DIFF
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -16,11 +16,12 @@ using System.Net;
 
 namespace Recurly {
     public class BaseClient {
-        private const string API_URL = "https://partner-api.recurly.com/";
         public string SiteId { get; }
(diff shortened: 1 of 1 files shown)


Recurly/BaseClient.cs
```diff
@@ -107,6 +108,16 @@ public class BaseClient {
             return resp;
         }
 
+        public void _SetApiUrl(string uri) {
+          if (System.Environment.GetEnvironmentVariable("RECURLY_INSECURE") == "true") {
+            Console.WriteLine("[SECURITY WARNING] _SetApiUrl is for testing only and not supported in production.");
```
aaron-suarez: I just realized that this `Console.WriteLine` is the same in both the `if` and the `else`. We should consider pulling it out of the logic and just print it first, then do the control flow.

PR #391, opened 2019-04-15
Allow targeting individual exceptions

bhelx:
This PR contains a breaking change which should be called out in the release. There are two main differences. 

1. Exception classes are now in the `Recurly.Errors` namespace. So change any instance of `Recurly.NetworkError` to `Recurly.Errors.NetworkError`.
2. ApiErrors no longer have the `ApiErrorType` enum at `ApiError#Type`. This has been changed to a string. We now throw a subclassed version of `ApiError` based on the type. You can fall back to the generic base class to catch all api errors. For example:

Before:

```csharp
try
{
    Account account = client.GetAccount(accountId);
}
catch (Recurly.ApiError ex)
{
    var apiErr = ex.Error;
    switch (apiErr.Type)
    {
        case Recurly.ApiErrorType.NotFound:
            Console.WriteLine("Resource Not Found");
        default:
            throw;
    }
}
```

After:

```csharp
try
{
  Account account = client.GetAccount(accountId);
}
catch (Recurly.Errors.NotFound ex)
{
  Console.WriteLine($"Resource Not Found: {ex.Error.Message}");
}
catch (Recurly.Errors.ApiError ex)
{
  Console.WriteLine($"Unexpected Recurly Error: {ex.Error.Message}");
}
```


README.md
```diff
@@ -98,33 +98,30 @@ try
...
+{
+  // Use base class ApiError to catch a generic error from the API
+  Console.WriteLine($"Unexpected Recurly Error: {ex.Error.Message}");
 }
 ```
 
-`Recurly.NetworkError`s don't come from Recurly's servers, but instead are triggered by some problem in
-related to the network. Depending on the context, you can often automatically retry these calls. GETs are always safe to retry but be careful about automatically re-trying any other call that might mutate state on the server side.
+`Recurly.Errors.NetworkError`s don't come from Recurly's servers, but instead are triggered by some problem in
+related to the network. Depending on the context, you can often automatically retry these calls.
```
aaron-suarez: "problem in related to the network"

PR #392, opened 2019-04-29
Dotnet Async Operations

bhelx:
This has a couple changes in it:

1. Adds async operations to the client
2. Adds async method to pager
3. Refactors pagination to be like the other clients (calling List* now does not make an http. The call is now made when consuming the pager)
4. Re-organizes specs to get > 80% coverage (will be more to come on this)

### DIFF
--- a/Recurly.Tests/Fixtures.cs
+++ b/Recurly.Tests/Fixtures.cs
@@ -1,0 +1,50 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Recurly.Tests
+{
+    public class MyResource : Resource
+    {
+        [JsonProperty("my_string")]
+        public string MyString { get; set; }
+
+        [JsonProperty("my_float")]
+        public float? MyFloat { get; set; }
+
+        [JsonProperty("my_int")]
+        public int? MyInt { get; set; }
+
+        [JsonProperty("my_date_time")]
+        public DateTime? MyDateTime { get; set; }
+
+        [JsonProperty("my_sub_resource")]
+        public MySubResource MySubResource { get; set; }
+
+        [JsonProperty("my_array_string")]
+        public List<string> MyArrayString { get; set; }
+
+        [JsonProperty("my_array_sub_resource")]
+        public List<MySubResource> MyArraySubResource { get; set; }
+    }
+
+    public class MySubResource : Resource {
+        [JsonProperty("my_string")]
+        public string MyString { get; set; }
+    }
+
+    public class MyResourceCreate : Request
+    {
+        [JsonProperty("my_string")]
+        public string MyString { get; set; }
+    }
+
+    public class MyApiError : Recurly.Errors.ApiError
+    {
+        public MyApiError() {}
+        public MyApiError(string message) : base(message) {}
+        public MyApiError(string message, Exception inner) : base(message, inner) {}
+    }
+
+}
--- a/Recurly.Tests/PagerTest.cs
+++ b/Recurly.Tests/PagerTest.cs
@@ -8,7 +8,7 @@ using Newtonsoft.Json;
 using RestSharp;
 using Moq;
 
-namespace Recurly.UnitTests
+namespace Recurly.Tests
 {
     public class PagerTest
     {
@@ -20,26 +20,24 @@ namespace Recurly.UnitTests
         [Fact]
         public void EnumerableTest()
         {
-            var pager = new Pager<MyResource>();
-            pager.HasMore = false;
-            pager.Data = new List<MyResource>() {
-                new MyResource() {
-                    MyString = "A String"
-                },
-                new MyResource() {
-                    MyString = "A String"
-                },
-                new MyResource() {
-                    MyString = "A String"
-                }
-            };
+            var pager = Pager<MyResource>.Build("/next", new Dictionary<string, object>{}, GetPagerSuccessClient());
 
             var i = 0;
             foreach(MyResource r in pager) {
-                Assert.Equal("A String", r.MyString);
+                if (i < 3)
+                {
+                    Assert.Equal("A page 1 String", r.MyString);
+                }
+                else
+                {
+                    Assert.Equal("A page 2 String", r.MyString);
+                }
                 i++;
             }
-            Assert.Equal(3, i);
+
+            // There should be 5 resources since
+            // there is a second page
+            Assert.Equal(5, i);
 
             // We don't allow resetting pager states right now
             Assert.Throws<NotImplementedException>(() => {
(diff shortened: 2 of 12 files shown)


Recurly.Tests/Fixtures.cs
```diff
@@ -0,0 +1,50 @@
...
+        public int? MyInt { get; set; }
+
+        [JsonProperty("my_date_time")]
+        public DateTime? MyDateTime { get; set; }
+
+        [JsonProperty("my_sub_resource")]
+        public MySubResource MySubResource { get; set; }
+
+        [JsonProperty("my_array_string")]
+        public List<string> MyArrayString { get; set; }
```
aaron-suarez: Does our API ever expose an array of anything besides strings and more resources? Like an array of date times or something crazy like that?

Recurly/BaseClient.cs
```diff
@@ -12,16 +15,18 @@
...
 
 namespace Recurly {
     public class BaseClient {
         public string SiteId { get; }
         private string ApiKey { get; }
-        public IRestClient RestClient { get; set; }
+        private const string ApiUrl = "https://partner-api.recurly.com/";
         public virtual string ApiVersion { get; protected set; }
 
-        private const string API_URL = "https://partner-api.recurly.com/";
```
aaron-suarez: Was the change from `API_URL` to `ApiUrl` just a style choice? What's the convention here?

Recurly/BaseClient.cs
```diff
@@ -45,11 +49,38 @@ public class BaseClient {
...
+            this.HandleResponse(resp);
+            return resp.Data;
+        }
+
+        public void _SetApiUrl(string uri) {
+          if (System.Environment.GetEnvironmentVariable("RECURLY_INSECURE") == "true") {
+            Console.WriteLine("[SECURITY WARNING] _SetApiUrl is for testing only and not supported in production.");
+            this.RestClient.BaseUrl = new Uri(uri);
+          } else {
+            Console.WriteLine("[SECURITY WARNING] _SetApiUrl is for testing only and not supported in production.");
```
aaron-suarez: We should pull out the duplicate `Console.WriteLine()` call to outside the `if/else` block since we're printing it either way

Recurly.Tests/Fixtures.cs
bhelx: not currently. Since this isn't really our code i feel like just testing a primitive is enough.

Recurly/BaseClient.cs
bhelx: This is the recommended coding convention in stylecop: https://github.com/StyleCop/StyleCop

Recurly/BaseClient.cs
bhelx: went ahead and changed.

Recurly/Pager.cs
```diff
@@ -40,16 +70,6 @@ public T Current
...
         }
 
-        private void FetchNextPage()
-        {
-            var response = RecurlyClient.MakeRequest<Pager<T>>(Method.GET, Next);
-            var pager = response.Data;
-            // this is brittle
-            this.Next = pager.Next;
-            this.Data = pager.Data;
-            this.HasMore = pager.HasMore;
```
aaron-suarez: I'm curious about the comment `this is brittle`. The logic is basically the same, but now it's inside the `Clone` method. Does putting the same code in a method make it less brittle? I don't see anything wrong with the way you have it now, I'm just trying to learn about the process here

Recurly/BaseClient.cs
```diff
@@ -73,8 +108,26 @@ public class BaseClient {
...
+                var deprecated = headers
+                    .Find(x => x.Name == "Recurly-Deprecated")
+                    .Value.ToString();
+                var sunset = headers
+                    .Find(x => x.Name == "Recurly-Sunset-Date")
+                    .Value.ToString();
+
+                if (deprecated.ToUpper() == "TRUE")
+                {
+                    Debug.WriteLine($"[recurly-client-net] WARNING: Your current API version \"${ApiVersion}\" is deprecated and will be sunset on ${sunset}");
```
aaron-suarez: I notice that with the security warning about using `_SetApiUrl`, we use `Console.WriteLine`. Do you think that people will miss this if it's only in the Debug log? Or perhaps this isn't important enough to warrant a Console write?

Recurly.Tests/PagerTest.cs
```diff
@@ -50,62 +48,46 @@ public void EnumerableTest()
             pager.Dispose();
```
aaron-suarez: Is this something we're going to build out later or is it something that's in the Interface but doesn't make sense for our context?

Recurly/Pager.cs
bhelx: I said it was brittle because it you have to keep it to date and it might be needed in multiple spots. It's now at least in one place. So less brittle but could maybe automatically look at all the properties and clone them over.

Recurly/BaseClient.cs
bhelx: I'm not sure yet. I haven't seen any document describing how to handle this kind of logging.

Recurly.Tests/PagerTest.cs
bhelx: This method must be present and callable to conform to the IEnumerable interface. I don't think it needs to do anything in our case though because we do not hold any resources.

Recurly/Pager.cs
aaron-suarez: Right, that makes sense. I agree that it would be a bit more robust to copy all the properties rather than pulling specific ones, but I also think it would be unlikely that this would cause problems in the future. Thanks for clarifying

Recurly/BaseClient.cs
aaron-suarez: I asked a question about Debug.WriteLine vs Console.WriteLine on another slack team I'm on. One person believes that Debug.WriteLine only shows up in the IDE. So if people set up their integration and then don't make any changes, they are less likely to see this in their logs. However, they might not be logging Console.WriteLine calls either, so no matter what we run the risk that no one will see this. I think it might make sense to switch to Console.WriteLine, but I don't think that will guarantee that anyone will see this.

aaron-suarez: Also, I can't find it in the documentation, but according to several sources I saw when I Googled it, Debug.WriteLine doesn't even get compiled into a Release build. I think we should just go ahead and change it to Console.WriteLine so it will print to stdout even in a Release build. Hopefully our users are logging calls to stdout, but that's beyond our control.

aaron-suarez: Someone recommended logging with `ILogger` https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.2

PR #395, opened 2019-05-13
Add a Timeout Configuration.

caleb-conner:
Add a timeout configuration to allow the user to overwrite the default timeout for an httpWebRequest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,12 +38,12 @@ The recommended way to configure the library is to statically initialize it:
 
 ```csharp
 // pageSize is optional and defaults to 200
-Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, pageSize);
+Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, privateKey, pageSize);
 ```
 
 ```csharp
 // requestTimeoutMilliseconds is optional and defaults to null
-Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, pageSize, requestTimeoutMilliseconds);
+Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, privateKey, pageSize, requestTimeoutMilliseconds);
 ```
 
 We recommend doing this for compatibility with newer .NET frameworks as well as security. We also recommend you encrypt
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, pageSize);
 ```
 
 ```csharp
-// requestTimeoutMilliseconds is potional and defaults to null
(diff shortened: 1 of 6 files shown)


aaron-suarez:
Thanks for your contribution @caleb-conner! We'll get this reviewed soon.

bhelx:
Very useful, thanks for taking the time to add this. Could we get it documented in the README somewhere? Perhaps as another subsection of `Configuration`: https://github.com/recurly/recurly-client-net#configuration

caleb-conner:
> Very useful, thanks for taking the time to add this. Could we get it documented in the README somewhere? Perhaps as another subsection of `Configuration`: https://github.com/recurly/recurly-client-net#configuration

I added the configuration section similar to how pageSize was documented. Let me know if you think more needs to be added.

README.md
```diff
@@ -41,6 +41,11 @@ The recommended way to configure the library is to statically initialize it:
 Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, pageSize);
 ```
 
+```csharp
+// requestTimeoutMilliseconds is potional and defaults to null
```
aaron-suarez: `potional` should be `optional`?

caleb-conner: Good catch fixed.

README.md
```diff
@@ -41,6 +41,11 @@ The recommended way to configure the library is to statically initialize it:
 Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, pageSize);
 ```
 
+```csharp
+// requestTimeoutMilliseconds is optional and defaults to null
+Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, pageSize, requestTimeoutMilliseconds);
```
aaron-suarez: I think I approved too quickly. Isn't this supposed to be
```C#
Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, privateKey, pageSize, requestTimeoutMilliseconds);
```

aaron-suarez: If you try it as the example lays it out, you get `error CS1503: Argument 3: cannot convert from 'int' to 'string'` because it thinks that the pageSize should be the string privateKey

caleb-conner: Good catch, I just pushed the correction. I also fixed it for the pageSize example since that one was wrong as well.

README.md
```diff
@@ -38,7 +38,12 @@ The recommended way to configure the library is to statically initialize it:
 
 ```csharp
 // pageSize is optional and defaults to 200
-Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, pageSize);
+Recurly.Configuration.SettingsManager.Initialize(apiKey, subdomain, privateKey, pageSize);
```
aaron-suarez: Thanks for updating this example!

Issue #411, opened 2019-06-10
Add 'protected' to constructors for RecurlyEntity


stweedie:
This is mainly so that we can properly mock Recurly objects for testing. Without providing a protected (or public) constructor, we are unable to mock several objects. Adding a protected constructor allows the following:

```c#
// change from "internal" -> "protected internal"
protected internal CouponRedemption()
{
}
```
```c#
public class TestRecurlyCouponRedemption: Recurly.CouponRedemption
{
	public TestRecurlyCouponRedemption(string couponCode, int amountInCents)
	{
		CouponCode = couponCode;
		TotalDiscountedInCents = amountInCents;
	}

	public new int TotalDiscountedInCents { get; set; }
	public new string CouponCode { get; set; }
	public new string Uuid { get; set; }

	public TestRecurlyCouponRedemption SetId(string id)
	{
		Uuid = id;
		return this;
	}
}
```

Therefore we can test our handling of coupon redemptions. 

Note: I have a branch on my personal fork that I can use to open a PR for this issue. I just wasn't sure if this idea would be accepted, but I know there are at least a few reasonable use cases where we would want to be able to mock some of the data structures coming from Recurly.

stweedie:
Alternately, and this might be more helpful in general, would you be willing to accept a PR to add interfaces for most recurly entities? Then we could provide our own implementation when needed (for mocking), but it would leave the entirety of the underlying implementation unchanged.

aaron-suarez:
Hi @stweedie! Thanks for bringing this up. I won't know for sure until we can review the code, but it seems like the original suggestion of making the properties `protected internal` should be just fine, especially if it helps for testing. If you think it would be better to add interfaces, that's an option we can look at as well, but it sounds like it would be more work to get to the same result, wouldn't it?

We'll be happy to review any PR you send to us, and the accessors you describe would likely be accepted. Go ahead and open it up and we can talk regarding the actual code changes. Thanks!

stweedie:
Alright I have opened a PR for the protected constructors.

I have since realized that there is some difficulty in mocking some of the instance methods, which is why I suggested the interface modification.

aaron-suarez:
Do you have a code sample of how you would mock it? I'd like to better understand what you're doing so I can verify the solution

stweedie:
so here would be an example, where we are trying to handle converting from a Recurly invoice to our own custom invoice model (where everything is stored as decimals instead of cents)

```c#
// TestRecurlyInvoice
[...]

PR #421, opened 2019-06-24
Extract interfaces

stweedie:
This is work related to issue #411.

Description of changes:
* Most classes that inherit from RecurlyEntity now have an interface. The only ones that don't have a related interface were private or internal to Recurly.
* Besides this, every reference to one of these classes (as a generic type or as a child property) now references the related interface. For example:
```c#
public class InvoiceCollection : RecurlyEntity, IInvoiceCollection
{
    public IInvoice ChargeInvoice { get; private set; }
    public IRecurlyList<IInvoice> CreditInvoices { get; private set; }
}
```
* Because of interface limitations on internal methods, I had to leverage internal extension methods and type checking to access / use the write xml methods. There is an alternative choice to use static methods and pass in the interface as an instance which I had to do in a few places, but some of the properties being written to XML were private.

I also was not able to run the unit tests. For some reason, VS2017 can't find a registered test runner, and installing suggested nuget packages gives SSL issues.


aaron-suarez:
Upon my initial look over the PR, I think it looks like a lot of breaking changes. I didn't realize when we discussed it before that implicit conversions weren't going to be possible. If merchants specify the type in their integration, like so:
```C#
Account account = Accounts.Get("x");
```
They'll now have to go in and change all of those from the class to the interface or it won't compile. I'm not sure how many people that would break, but it seems like it might cause problems for those that aren't necessarily trying to create a lot of tests around the library.

We'll have to talk internally about whether this amount of breaking changes is going to be something we want to do. My initial thought is to not break V2 so hard because we're going to want to get people onto V3 the next time they have to make a bunch of breaking changes to their integrations.

There's no way to have these interfaces while still preserving the old classes, is there?

stweedie:
I can definitely go back and revert the static calls (for instance, Accounts.Get would return an Account not an IAccount)

Unfortunately, properties on the classes themselves would still have to be interfaces, so there would still be the potential for breaking changes.

As an example,
```C#
public interface IAdjustment : IRecurlyEntity
{
    IShippingAddress ShippingAddress { get; }
}

public class Adjustment : RecurlyEntity, IAdjustment
{
    IShippingAddress ShippingAddress { get; private set; }
}
```

Adjustment.ShippingAddress would still have to be type ```IShippingAddress```.

I'm not sure if this will sufficiently prevent the breaking changes you describe, though

aaron-suarez:
Out of curiosity @stweedie, did you use a tool to make all these conversions or did you manually change everything required for this PR?

stweedie:
I used the 'Extract Interface' tool that is in visual studio, but I did have to manually find all references to the concrete classes and replace them with interfaces.

stweedie:
Alright I have reverted the changes to the static methods. Static methods (such as ```Adjustments.Get()``` will return the concrete type rather than an interface.

Should I go ahead and unwind the IRecurlyList stuff?

aaron-suarez:
I'm glad to hear that the V3 client looks like it will help resolve your mocking/testing issues. We expect to go GA by the end of the year, but you can start using it now as part of the early access program. Check out https://dev.recurly.com/page/recurly-v3-api-early-access and also feel free to send an email to dx@recurly.com to get help getting set up.

That being said, I think I'm going to hand off this conversation to @bhelx who can help guide you further on what he would like to see to get something merged.

bhelx:
Hey @stweedie, @aaron-suarez passed this on to me. I'm gonna take a look at it soon. We are heavily focused on preparing for V3 and I'm very hesitant to make any unnecessary changes to this library as it's been the hardest for us to support. This seems benign but the fact that it's 91 files does worry me. It's a lot of surface area to make sure is correct.

In the mean time, we'd love your feedback on V3. Since we are still in beta, I'm much more open to changing the interface.

bhelx:
@stweedie I thought about this a bit, although it's a good change I think it's just too many files changed and it increases the cognitive load in order to make a change. I'm trying to limit the amount of changes here to necessary changes only so we can focus on the next version of the client. I apologize for closing this late in the process. I should have chimed in earlier that this might be a problem.

Feel free to email us with questions or concerns. Like I said before, would love to get your opinions on V3 which is in a much more fluid position at the moment. Email: dx@recurly.com

PR #426, opened 2019-07-09
Only send Address object if Address properties were changed

aaron-suarez:
When updating an invoice, more XML elements are written than what was actually changed. This can cause problems if, for example, a site has taxes enabled and an invoice is previously taxed (meaning that the address cannot be updated).

Suppose you want to update the invoice PO number. The XML should look like this:
```xml
<?xml version="1.0" encoding="utf-8"?>
<invoice>
  <po_number>1-I1</po_number>
</invoice> 
```

Currently, an Address object is also sent with the invoice update, like so:
```xml
<?xml version="1.0" encoding="utf-8"?>
<invoice>
  <address>
    <first_name>Aaron</first_name>
    <last_name>Du Monde</last_name>
    <name_on_account />
    <company />
    <address1>900 Camp St.</address1>
    <address2 />
    <city>New Orleans</city>
    <state>LA</state>
    <zip>70115</zip>
    <country>US</country>
    <phone />
  </address>
  <customer_notes />
  <terms_and_conditions />
  <vat_reverse_charge_notes />
  <gateway_code />
  <po_number>1-I1</po_number>
</invoice>
```

This PR also removes the `customer_notes`, `terms_and_conditions`, `vat_reverse_charge_notes`, `gateway_code` and `po_number` from the request unless they are explicitly set as empty strings. The API will sometimes return empty strings for these values when they aren't explicitly set, but now if you would like the request to avoid including those empty strings, you can set them to null.

Script:
```c#
// Assumes invoice 3256 exists on the site
var invoice = Invoices.Get(3256);
invoice.PoNumber= "1-I1";
invoice.CustomerNotes = null;
try
{
    invoice.Update();
}
catch(ValidationException e)
{
    Console.WriteLine(e);
    foreach (var err in e.Errors) {
        Console.WriteLine(err);
    }
}
```

### DIFF
--- a/Library/Invoice.cs
+++ b/Library/Invoice.cs
@@ -108,6 +108,9 @@ namespace Recurly
         }
         private Address _address;
 
+        // Preserve original address in case the user wants to change a field in the address
+        private Address _referenceAddress;
+
         public ShippingAddress ShippingAddress { get; private set; }
 
         /// <summary>
@@ -586,6 +589,7 @@ namespace Recurly
 
                     case "address":
                         Address = new Address(reader);
+                        _referenceAddress = (Address) Address.Clone();
                         break;
 
                     case "shipping_address":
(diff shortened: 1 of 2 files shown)


Library/Invoice.cs
```diff
@@ -665,12 +683,18 @@ internal void WriteUpdateXml(XmlTextWriter xmlWriter)
...
-            Address.WriteXml(xmlWriter);
-            xmlWriter.WriteElementString("customer_notes", CustomerNotes);
-            xmlWriter.WriteElementString("terms_and_conditions", TermsAndConditions);
-            xmlWriter.WriteElementString("vat_reverse_charge_notes", VatReverseChargeNotes);
-            xmlWriter.WriteElementString("gateway_code", GatewayCode);
-            xmlWriter.WriteElementString("po_number", PoNumber);
+            if (!Address.Equals(_referenceAddress)) Address.WriteXml(xmlWriter);
+
+            if (CustomerNotes != null)
+                xmlWriter.WriteElementString("customer_notes", CustomerNotes);
```
aaron-suarez: This will technically be a breaking change, but I think it will add value along the same lines as the conversation here https://github.com/recurly/recurly-client-net/pull/368#discussion_r246890848

We'll go ahead and bump the minor version when we release this.

Library/Invoice.cs
```diff
@@ -586,6 +589,21 @@ internal void ReadXml(XmlTextReader reader, string nodeName)
 
                     case "address":
                         Address = new Address(reader);
+                        _referenceAddress = new Address()
```
bhelx: This seems pretty fragile, could you not clone this object?

aaron-suarez: I didn't see a built-in `DeepCopy` method for this class, but I can do a bit more research to determine if there is a more robust option for cloning the Address object

aaron-suarez: Added a `Address.Clone()` method, which uses `MemberwiseClone` under the hood. Less fragile?

PR #431, opened 2019-07-16
Bump to version 1.16.0

aaron-suarez:
1.16.0 (stable) / 2019-07-16
===============

* Fix timeoutMilliseconds when using config file [PR](https://github.com/recurly/recurly-client-net/pull/420)
* Only send Address object if Address properties were changed [PR](https://github.com/recurly/recurly-client-net/pull/426)
* Add missing properties to XML serializer for Adjustment class  [PR](https://github.com/recurly/recurly-client-net/pull/427)

### Upgrade Notes
This release contains one breaking change:

When updating an Invoice, the embedded Address object is only sent to the API if explicitly set on the object, and the following fields require an empty string to clear values:
- CustomerNotes
- TermsAndConditions
- VatReverseChargeNotes
- GatewayCode
- PoNumber

In the past, if you were to leave a value such as `CustomerNotes` null, it would nullify it via the API.
Now, you must explicitly nullify it by setting it to empty string `""`. See [this conversation](https://github.com/recurly/recurly-client-net/pull/368#discussion_r246890848) for an example.

### DIFF
--- a/Test/BillingInfoTest.cs
+++ b/Test/BillingInfoTest.cs
@@ -78,7 +78,7 @@ namespace Recurly.Test
             {
                 billingInfo.Update();
             }
-            catch (NotFoundException exception)
+            catch (ValidationException exception)
             {
                 threw = true;
                 exception.Errors[0].Symbol.Should().Be("token_invalid");
--- a/History.md
+++ b/History.md
@@ -1,6 +1,26 @@
 Unreleased
 ===============
 
+1.16.0 (stable) / 2019-07-16
+===============
+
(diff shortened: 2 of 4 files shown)


Test/BillingInfoTest.cs
```diff
@@ -78,7 +78,7 @@ public void UpdateBillingInfoWithToken()
             {
                 billingInfo.Update();
             }
-            catch (NotFoundException exception)
+            catch (ValidationException exception)
```
aaron-suarez: This should have been changed in the 1.15.8 release, but it wasn't so I went ahead and fixed it now

PR #432, opened 2019-07-18
Fix Pager empty page bug

bhelx:
In consulting with @aaron-suarez on the Java pager, I realized there wasn't a test for an empty page so I added it and it failed. This resulted me in moving some things around to better handle this case.

### Test the case:

```csharp
int count = 0;

var accountId = Utils.CreateAccount(client).Id;
// Try with shipping addresses and without
//Utils.CreateShippingAddress(client, accountId);

var addresses = client.ListShippingAddresses(accountId);
foreach(ShippingAddress address in addresses)
{
  count++;
}

Console.WriteLine("For-each count:");
// should be 0 (or 1 if CreateShippingAddress uncommented)
Console.WriteLine(count);
```

### Regression Test

```csharp
int count = 0;

var accounts = client.ListAccounts(null, 200);
foreach(Account account in accounts)
{
     count++;
}
Console.WriteLine("For-each count:");
Console.WriteLine(count); // Get the expected count from your recurly dashboard

count = 0;
var accounts = client.ListAccounts(null, 200);
while(accounts.HasMore)
{
  Console.WriteLine("Fetching next page...");
  accounts.FetchNextPage();
  foreach(Account a in accounts.Data)
  {
    count++;
   }
}

Console.WriteLine("Each-Page Count:");
Console.WriteLine(count); // Get the expected count from your recurly dashboard
```

### DIFF
--- a/Recurly.Tests/PagerTest.cs
+++ b/Recurly.Tests/PagerTest.cs
@@ -19,6 +19,21 @@ namespace Recurly.Tests
         }
 
         [Fact]
+        public void EmptyEnumerableTest()
+        {
+            var pager = Pager<MyResource>.Build("/next", new Dictionary<string, object> { }, GetEmptyPagerClient());
+
+            var i = 0;
+            foreach (MyResource r in pager)
+            {
+                Assert.True(false, "Should not be iterating anything if response is empty");
+            }
+
+            // There should be 0 resources
+            Assert.Equal(0, i);
+        }
+
+        [Fact]
         public void EnumerableTest()
         {
             var pager = Pager<MyResource>.Build("/next", new Dictionary<string, object> { }, GetPagerSuccessClient());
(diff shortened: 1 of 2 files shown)


Recurly.Tests/PagerTest.cs
```diff
@@ -70,11 +85,11 @@ public void EnumerablePagesTest()
                 }
                 if (page == 1)
                 {
-                    Assert.Equal(count, 3);
+                    Assert.Equal(3, count);
                 }
                 else if (page == 2)
                 {
-                    Assert.Equal(count, 2);
+                    Assert.Equal(2, count);
```
bhelx: This fix some warnings. The literal should be first.

Recurly/Pager.cs
```diff
@@ -97,7 +92,12 @@ IEnumerator IEnumerable.GetEnumerator()
 
         public bool MoveNext()
         {
-            return _index < Data.Count || HasMore;
+            if (HasMore && (Data == null || _index >= Data.Count))
```
bhelx: Gonna add some comments to this code

PR #456, opened 2019-10-25
[WIP] Autoinitialize List Properties

### LINKED ISSUE #455: List properties on Requests should be auto-initialized
Reported by @BenGoldberg in #445 

> I noticed the List properties in the Create objects don't come pre-initialized. For example, for the SubscriptionCreate object, there is an AddOns property to add add ons to the request. Right now I have to create a new list and then add the SubscriptionAddOnCreate objects inside of it. Would it make sense to initialize these list properties when creating the objects so you would only have to add to them when creating the requests? I ran into a couple unexpected NPE because of this.

bhelx:
Resolves #455 

This allows you to `Add` to List properties in Request objects without initializing them, reducing chance of NPEs. I'm still looking at how this affects serialzation / deserialization. We might want to avoid this if it affects that behavior. Example:

```csharp
var accountReq = new AccountCreate()
{
    Code = accountCode,
};
accountReq.ShippingAddresses.Add(new ShippingAddressCreate() {
    FirstName = "Benjamin",
    LastName = "Du Monde",
    Street1 = "123 Tchoupitoulas St",
    City = "New Orleans",
    Region = "LA",
    PostalCode = "70130",
    Country = "US"
});
Account account = client.CreateAccount(accountReq);
```

### DIFF
--- a/Recurly/Client.cs
+++ b/Recurly/Client.cs
@@ -1356,30 +1356,32 @@ namespace Recurly
         /// Collect a pending or past due, automatic invoice <see href="https://developers.recurly.com/api/v2019-10-10#operation/collect_invoice">collect_invoice api documentation</see>
         /// </summary>
         /// <param name="invoiceId">Invoice ID or number (use prefix: `number-`, e.g. `number-1000`).</param>
+        /// <param name="body">The body of the request.</param>
         /// <returns>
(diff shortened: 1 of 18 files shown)


bhelx:
Closing until we pick it up again.

PR #457, opened 2019-10-31
Method to download a PDF invoice

oktaras:
Hello!

During development, we needed a method to download a PDF invoice. Something like in the previous version of your library. We could not find a solution in the current version. Then we turned to support to explain to us how to do this. We managed to implement this, and it works. We would like this method to be available in the library when we connect it as a NuGet package. I hope for a quick solution to this issue. 

Thank you for your attention!

### DIFF
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -70,6 +70,24 @@ namespace Recurly
             return resp.Data;
         }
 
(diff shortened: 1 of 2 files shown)


bhelx:
@oktaras Thanks so much for taking the time to do this! This is great work. The `Client.cs` file isn't meant to be edited by hand as it's generated from the OpenAPI schema. This is not a big problem, we just need some time implement pdf support in the generator. Please give us some time to do this before we merge and release.

douglasmiller:
@oktaras, I wanted to let you know that we have updated the code generation process to include the `GetInvoicePdf` method. We couldn't make direct usage of your PR but it was used as a guide and cut down on a lot of time. Thanks again for contributing and prompting this update!

The 3.1.0 version of the package has been published and is available for usage!

PR #467, opened 2019-11-27
POC enum support

douglasmiller:


Recurly/Resources/Account.cs
```diff
@@ -17,94 +20,117 @@ public class Account : Resource
...
         [JsonProperty("last_name")]
+
         public string LastName { get; set; }
 
         /// <value>The UUID of the parent account associated with this account.</value>
         [JsonProperty("parent_account_id")]
+
         public string ParentAccountId { get; set; }
 
-        /// <value>Used to determine the language and locale of emails sent on behalf of the merchant to the customer.</value>
```
bhelx: We need to preserve these docstrings if we can

Recurly/Resources/AccountAcquisitionCost.cs
```diff
@@ -17,10 +20,12 @@ public class AccountAcquisitionCost : Request
 
         /// <value>The amount of the corresponding currency used to acquire the account.</value>
         [JsonProperty("amount")]
+
```
bhelx: I think these probabaly should sit right on top of their attribute they are referring to

douglasmiller: I fixed this recently

Recurly/Resources/Account.cs
douglasmiller: Fixed and apparently pulled in some new ones as well!

Recurly/Resources/AddOn.cs
```diff
@@ -55,9 +55,10 @@ public class AddOn : Resource
         [JsonProperty("plan_id")]
         public string PlanId { get; set; }
 
-        /// <value>Add-ons can be either active or inactive.</value>
+        /// <value>Active state can be either active or inactive.</value>
```
douglasmiller: Changes like this happened because the `ActiveState` enum is used in multiple places so I adjusted the description to be more generic because it's defined on the enum schema.

douglasmiller:
This was merged into master via https://github.com/recurly/recurly-client-net/pull/471

PR #476, opened 2020-02-05
Add external_sku to Adjustment

joannasese:
To fetch `external_sku`:
```csharp
var adjustment = account.NewAdjustment("USD", 5000);     
adjustment.ItemCode = 'item_code';                
adjustment.StartDate = new DateTime(2019, 6, 11);
adjustment.EndDate = new DateTime(2045, 7, 11);
adjustment.Create();
                
System.Console.WriteLine("Adjustment created: " + adjustment.ExternalSku);
```

### DIFF
--- a/Library/Adjustment.cs
+++ b/Library/Adjustment.cs
@@ -39,6 +39,7 @@ namespace Recurly
         public string AccountingCode { get; set; }
         public string ProductCode { get; set; }
         public string ItemCode { get; set; }
+        public string ExternalSku { get; set; }
         public string Origin { get; set; }
         public int UnitAmountInCents { get; set; }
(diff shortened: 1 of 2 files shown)


Library/Adjustment.cs
```diff
@@ -39,6 +39,7 @@ public enum RevenueSchedule : short
         public string AccountingCode { get; set; }
         public string ProductCode { get; set; }
         public string ItemCode { get; set; }
+        public string ExternalSku{ get; set; }
```
bhelx: nit: need a space here before the `{`

joannasese: Fixed!

PR #480, opened 2020-02-11
Add ConvertTrial() to Subscription

joannasese:
To convert trial to paid subscription with 3ds token:
```csharp
subscription.ConvertTrial("token")
```

To convert trial for non-3ds where subscription has billing info:
```csharp
subscription.ConvertTrial()
```

To convert trial to paid subscription with transaction type:
```csharp
subscription.ConvertTrialMoto()
```

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,14 @@
 Unreleased
 ===============
 
+1.17.1 (stable) / 2020-02-18
+===============
+
+- Add 'capture' and 'cancel' to Purchase [PR](https://github.com/recurly/recurly-client-net/pull/474)
+- Repair broken tests [PR](https://github.com/recurly/recurly-client-net/pull/475)
+- Add external_sku to Adjustment [PR](https://github.com/recurly/recurly-client-net/pull/476)
+- Request will include item_code if ItemCode has a value [PR](https://github.com/recurly/recurly-client-net/pull/477)
+
 1.17.0 (stable) / 2019-11-21
 ===============
 
--- a/Library/BillingInfo.cs
+++ b/Library/BillingInfo.cs
@@ -178,7 +178,7 @@ namespace Recurly
             ReadXml(reader);
         }
 
-        private BillingInfo()
+        internal BillingInfo()
         {
         }
 
--- a/Library/Configuration/Settings.cs
+++ b/Library/Configuration/Settings.cs
@@ -78,7 +78,7 @@ namespace Recurly.Configuration
         }
 
         protected const string RecurlyServerUri = "https://{0}.recurly.com/v2{1}";
-        public const string RecurlyApiVersion = "2.24";
+        public const string RecurlyApiVersion = "2.25";
         public const string ValidDomain = ".recurly.com";
 
         // static, unlikely to change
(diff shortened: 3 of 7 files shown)


bhelx:
I think this could go something like this:

```csharp
        private void ConvertTrial(Subscription request)
        {
            var url = UrlPrefix + Uri.EscapeDataString(Uuid) + "/convert_trial";
            if (request == null)
            {
                Client.Instance.PerformRequest(Client.HttpRequestMethod.Put,
                    url,
                    ReadXml);
            }
            else
            {
                Client.Instance.PerformRequest(Client.HttpRequestMethod.Put,
                    url,
                    request.WriteSubscriptionXml, 
                    ReadXml);
            }
        }
        
        public void ConvertTrialMoto() {
            var request = new Subscription()
            {
                TransactionType = "moto"
            };  
            ConvertTrial(request);
        }

        public void ConvertTrial(string threeDSecureActionResultTokenId = null) {
            // by default there is no body
            Subscription request = null;

            // if there is a 3dsecure token, let's build a request body for it
            if (threeDSecureActionResultTokenId != null)
            {
                request = new Subscription()
                {
                    Account = new Account()
                    {
                        BillingInfo = new BillingInfo()
                        {
                            ThreeDSecureActionResultTokenId = threeDSecureActionResultTokenId
                        }
                    }
                };
            }

            ConvertTrial(request);
        }
```

You'd also need to add an internal setter for Subscription#Account. I added this to line `57`: 
```csharp
            internal set { _account = value; }
```

That gives you 3 ways to call it:

```csharp
// For non-3ds where billing info is already there to be converted
subscription.ConvertTrial();

// Optionally pass in a token if we need 3ds auth
subscription.ConvertTrial(token);

// If we want to explicitly skip the authorization step (like for a MOTO order)
subscription.ConvertTrialMoto();
```

Unfortunately though, we can't just use the `WriteSubscriptionXml` serializer and the second and third calls won't work out of the box. Although it mostly works, it creates a bunch of junk which the server will not like. Example when using a token:

```xml
<subscription>
  <plan_code />
  <account>
    <account_code />
    <billing_info>
      <three_d_secure_action_result_token_id>token</three_d_secure_action_result_token_id>
    </billing_info>
  </account>
  <currency />
  <customer_notes />
  <terms_and_conditions />
  <vat_reverse_charge_notes />
  <po_number />
  <subscription_add_ons />
  <quantity>0</quantity>
</subscription>
```

So I think you'll want to write your own `writeXml*` serializer in `Subscription` which only renders out the appropriate xml like so:

```xml
<subscription>
  <account>
    <billing_info>
      <three_d_secure_action_result_token_id>token</three_d_secure_action_result_token_id>
    </billing_info>
  </account>
</subscription>
```

Library/Adjustment.cs
```diff
@@ -299,14 +299,15 @@ internal void WriteXml(XmlTextWriter xmlWriter, bool embedded = false)
             xmlWriter.WriteStartElement("adjustment"); // Start: adjustment
             xmlWriter.WriteElementString("unit_amount_in_cents", UnitAmountInCents.AsString());
             xmlWriter.WriteElementString("quantity", Quantity.AsString());
-            xmlWriter.WriteElementString("item_code", ItemCode);
```
joannasese: This is from https://github.com/recurly/recurly-client-net/pull/477 and somehow ended up here bc I didn't git right.

PR #481, opened 2020-02-12
Convert Floats to Decimals with Monetary Amounts

jamesbar2:
I've opened a PR to show what the types should look like when generated to C#.  Floating-point types are not accurate for monetary amounts.  From [Microsoft](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/types#the-decimal-type): 
> The decimal type is a 128-bit data type suitable for financial and monetary calculations. The decimal type can represent values ranging from 1.0 * 10^-28 to approximately 7.9 * 10^28, with 28-29 significant digits.

Additional information on Floating Points can be found here: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types

You may be able to patch the OpenApi specification, but it officially only supports `number` type up to `double` ([src](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#data-types)).

Additionally, I've patched the tests to work locally regardless of the timezone your machine is in.

### DIFF
--- a/Recurly.Tests/UtilsTest.cs
+++ b/Recurly.Tests/UtilsTest.cs
@@ -26,10 +26,10 @@ namespace Recurly.Tests
             d.Add("trueBool", true);
             d.Add("falseBool", false);
             d.Add("int", 123);
-            d.Add("float", 123.456);
-            d.Add("date", new DateTime(2020, 1, 1));
+            d.Add("decimal", 123.456m);
+            d.Add("date", new DateTime(2020, 1, 1, 0, 0, 0, DateTimeKind.Utc));
             var result = Utils.QueryString(d);
-            Assert.Equal("?trueBool=true&falseBool=false&int=123&float=123.456&date=2020-01-01T00:00:00.000Z", result);
+            Assert.Equal("?trueBool=true&falseBool=false&int=123&decimal=123.456&date=2020-01-01T00:00:00.000Z", result);
         }
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ msbuild.wrn
 
 coverage.*
 coverage_report/
+
+packages/
+.idea/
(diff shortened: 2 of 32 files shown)


bhelx:
@jamesbar2 really appreciate your input here. Would you mind breaking the test change into a different PR? I think we can merge that right away.

As for the "money as float" issue, it's something we are aware of and we hope to address in the next major version across all the libraries but we've yet to settle on a strategy. The solution would be to go back to integers and cents, or to do use a `decimal` or custom Money type depending on the language. We can extend OpenAPI to support this.

jamesbar2:
@bhelx Just the DateTime fix, I imagine?

bhelx:
@jamesbar2 correct! thanks!

bhelx:
I'm going to further discuss the float change with my team. We'll leave this ticket open in the mean time.

bhelx:
@jamesbar2 Just wanted to chime in and let you know we did not forget about this `decimal` change. Since this change could break compilation for some people, we'd like to publish it in the next major version (v4). That is on schedule for the start of Q3 and this change will be released with it.

Recurly.Tests/UtilsTest.cs
```diff
@@ -26,10 +26,10 @@ public void QueryString()
             d.Add("trueBool", true);
             d.Add("falseBool", false);
             d.Add("int", 123);
-            d.Add("float", 123.456);
-            d.Add("date", new DateTime(2020, 1, 1));
+            d.Add("decimal", 123.456m);
+            d.Add("date", new DateTime(2020, 1, 1, 0, 0, 0, DateTimeKind.Utc));
```
bhelx, line 30: Why is this change needed? is this from the original author's commit?

bhelx:
@jamesbar2 closing this to complete it in #562. @joannasese preserved your original commit for your credit. This will go out in the next major release (targeting September)

PR #482, opened 2020-02-12
Fix a DateTime Unit Tests when not in UTC

jamesbar2:
- Fixes a unit test that fails locally when not in UTC.
- Adds a float specification to the object added to the Dictionary.
- .gitignore updates to not checking build packages or Rider IDE files.


bhelx:
Thanks again!

PR #489, opened 2020-03-09
Enable item-backed add-ons

joannasese:
To create item-backed add-on:
```c#
var addon = plan.NewAddOn("item_code");
addon.UnitAmountInCents.Add("USD", 200);
addon.Create();
```

To get an item-backed-add-on:
```c#
plan.GetAddOn(item.ItemCode);
```

To update item-backed add-on:
```c#
addon.UnitAmountInCents["USD"] = 200;
addon.Update();
```

When add-on has an associated item, only the following attributes are updatable:
```c#
unit_amount_in_cents
default_quantity
display_quantity_on_hosted_page
optional
```

### DIFF
--- a/Library/AddOn.cs
+++ b/Library/AddOn.cs
@@ -14,6 +14,7 @@ namespace Recurly
 
         public string PlanCode { get; set; }
         public string AddOnCode { get; set; }
+        public string ItemCode {get; set; }
         public string Name { get; set; }
         public int? DefaultQuantity { get; set; }
         public bool? DisplayQuantityOnHostedPage { get; set; }
@@ -23,9 +24,11 @@ namespace Recurly
         public long? MeasuredUnitId { get; set; }
         public Type? AddOnType { get; set; }
         public Usage.Type? UsageType { get; set; }
+        public float? UsagePercentage {get; set; }
         public DateTime CreatedAt { get; private set; }
         public DateTime UpdatedAt { get; private set; }
         public Adjustment.RevenueSchedule? RevenueScheduleType { get; set; }
+        public string ItemState { get; set; }
 
         private Dictionary<string, int> _unitAmountInCents;
         /// <summary>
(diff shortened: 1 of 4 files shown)


Library/AddOn.cs
```diff
@@ -176,13 +204,27 @@ internal override void ReadXml(XmlTextReader reader)
             }
         }
 
-        internal override void WriteXml(XmlTextWriter xmlWriter)
+        internal void WriteItemBackedAddOnXml(XmlTextWriter xmlWriter)
         {
             xmlWriter.WriteStartElement("add_on");
+
```
joannasese: Given the following update request:
```
addon.UnitAmountInCents["USD"] = 400;
addon.Update(item.ItemCode);
```

request xml is always 

```
<add_on>
  <unit_amount_in_cents>
    <USD>400</USD>
  </unit_amount_in_cents>
  <default_quantity>1</default_quantity>
  <display_quantity_on_hosted_page>false</display_quantity_on_hosted_page>
  <optional>true</optional>
</add_on>
```
when it should be
```
<add_on>
  <unit_amount_in_cents>
    <USD>400</USD>
  </unit_amount_in_cents>
</add_on>
```

Library/AddOn.cs
```diff
@@ -80,6 +88,34 @@ public void Update()
                 ReadXml);
         }
 
+        /// <summary>
+        /// Update an existing item-backed add on in Recurly
+        /// </summary>
+            public void UpdateItemBackedAddOn(
```
joannasese: A clumsy implementation. Open to other ways to approach this.

joannasese: A clumsy implementation. Open to other ways to approach this.

Library/Plan.cs
```diff
@@ -171,6 +171,12 @@ public AddOn NewAddOn(string addOnCode, string name)
             return a;
         }
 
+        public AddOn NewAddOn(string itemCode)
+        {
+            var a = new AddOn(PlanCode, itemCode);
+            return a;
```
bhelx, line 177: no need for a variable here?
```
return new AddOn(PlanCode, itemCode);
```

joannasese: That's a good point.

PR #490, opened 2020-03-13
Add IBAN attribute to BillingInfo

joannasese:
This feature is to allow support of the IBAN field in our V2 API for the applicable endpoints. This field is most similar to banking account information or credit card information, and is used to make payments for SEPA transactions.

IBAN numbers can be set in all endpoints that support BillingInfo:

PUT "/v2/accounts/{account_id}/billing_info"
PUT "/v2/accounts/{account_id}"
POST "/v2/accounts"
POST "/v2/subscriptions"
POST "/v2/purchases"
POST "/v2/purchases/preview"

A sample request payload is provided below:

```
<billing_info>
  <iban>IBANNUMBERHERE</iban>
  <name_on_account>Verena Example</name_on_account>
  <!-- address fields... ->
</billing_info>
```

### DIFF
--- a/Test/BillingInfoTest.cs
+++ b/Test/BillingInfoTest.cs
@@ -88,6 +88,29 @@ namespace Recurly.Test
         }
 
         [RecurlyFact(TestEnvironment.Type.Integration)]
+        public void CreateBillingInfoWithIban()
+        {
+            var account2 = CreateNewAccount();
+            var ibanInfo = new BillingInfo(account2)
+            {
+                NameOnAccount = "Iban account name",
+                Iban = "FR1420041010050500013M02606",
+                Address1 = "123 Test St",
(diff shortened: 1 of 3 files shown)


Test/BillingInfoTest.cs
```diff
@@ -87,6 +87,29 @@ public void UpdateBillingInfoWithToken()
...
+                Iban = "FR1420041010050500013M02606",
+                Address1 = "123 Test St",
+                Address2 = "The Test Cut",
+                City = "San Francisco",
+                PostalCode = "94105",
+                State = "CA",
+                Country = "US",
+            };
+            try {
+                ibanInfo.Create();
```
joannasese: This test actually does not pass because `card_type=invalid` in this scenario.

PR #491, opened 2020-03-17
Adding #First and #Count methods to Pager

douglasmiller:
Adds the `First` and `Count` methods to the `Pager` class.

The `First` method will perform a request to the API with the `limit` set to 1 and will only return the first result from the server.

The `Count` method will perform a `HEAD` request to the API for the appropriate endpoint. The value of the response header, `recurly-total-records`, will be returned. This is the total number of records that will be returned by the pager.

### DIFF
--- a/Recurly.Tests/PagerTest.cs
+++ b/Recurly.Tests/PagerTest.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Collections;
 using System.Collections.Generic;
 using Xunit;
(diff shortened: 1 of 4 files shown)


Recurly.Tests/PagerTest.cs
```diff
@@ -168,5 +200,74 @@ private Recurly.Client GetEmptyPagerClient()
...
+            pageResponse.Setup(_ => _.Headers).Returns(new List<Parameter> { });
+            pageResponse.Setup(_ => _.Data).Returns(page);
+
+            var mockIRestClient = new Mock<IRestClient>();
+            var myParams = new List<Parameter> {
+                new RestSharp.Parameter("limit", "1", ParameterType.QueryString),
+                new RestSharp.Parameter("a", "1", ParameterType.QueryString),
+            };
+
+            // TODO: Find a better way to handle this specifically and the concept at a whole
```
bhelx: For anyone looking, we are going to follow up fixing the request end of the specs in this library. It does a good job of mocking the responses but a bad job of verifying the requests. In the follow up we will address the problem as a whole.

Recurly.Tests/PagerTest.cs
```diff
@@ -102,6 +103,36 @@ public void EnumerablePagesTest()
             Assert.Equal(5, total);
         }
 
+        [Fact]
+        public void PagerFirstTest()
+        {
+            Dictionary<string, object> queryParams = new Dictionary<string, object> {
```
bhelx, line 109: I think the left-hand-side can use `var`. Unlike Java, csharp should be smart enough to know the left-hand side type given the right hand's declaration.

PR #503, opened 2020-04-01
Allow tiered add-ons to be updated

joannasese:
- Create separate serializer that excludes `unit_amount_in_cents` in request body

To update tiered add-on:
```c#
addon.UnitAmountInCents["USD"] = 678;
addon.Update();
```

### DIFF
--- a/Library/AddOn.cs
+++ b/Library/AddOn.cs
@@ -86,17 +86,10 @@ namespace Recurly
         /// </summary>
         public void Update()
         {
-            if (this.TierType == "tiered") {
-                Client.Instance.PerformRequest(Client.HttpRequestMethod.Put,
-                    UrlPrefix + Uri.EscapeDataString(PlanCode) + UrlPostfix + Uri.EscapeDataString(AddOnCode),
-                    WriteTieredUpdateXml,
-                    ReadXml); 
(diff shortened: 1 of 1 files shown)


Library/AddOn.cs
```diff
@@ -231,6 +238,41 @@ internal override void WriteXml(XmlTextWriter xmlWriter)
             xmlWriter.WriteEndElement();
         }
 
+        internal void WriteTieredUpdateXml(XmlTextWriter xmlWriter) {
```
bhelx, line 241: Does it make sense to have a separate method for this when most of the values are the same? Worried it might result in people applying serializer updates to the wrong method or only one of them.

joannasese: Alternatively, I could alter `WriteXml()` to include the conditional below:
```c#
if (this.TierType == "flat")
  xmlWriter.WriteIfCollectionHasAny("unit_amount_in_cents", UnitAmountInCents, pair => pair.Key,
  pair => pair.Value.AsString());
```
That minor change rids us of the error `currencies cannot be used with tiers` that gets thrown when invoking `WriteXml()` in `Update()` in instances where `tier_type != flat`.

PR #510, opened 2020-04-27
Encode forward slashes in URL parameters

douglasmiller:
Encoding url path parameters


douglasmiller:
I refactored the tests to facilitate code reuse when writing a new test that needed to validate a component of the request

PR #515, opened 2020-05-22
Update serializer for item-backed add-ons to accommodate QBP

joannasese:
**For item-backed QBP add-ons:**
This PR changes the [serializer for item-backed add-ons](https://ci-master.dev-inf.recurly.net/blue/organizations/jenkins/recurly-github%2Frecurly-app/detail/PR-15610-head/5/pipeline) to include quantity-based pricing tiers. 

**For general QBP add-ons:**
A previous [PR](https://github.com/recurly/recurly-client-dotnet/pull/503/files) considered cases where `TierType == flat`. This allows us to update add-ons by only including the `unit_amount_in_cents ` element in the absence of QBP. Otherwise, `unit_amount_in_cents` is [included within `Tier`](https://github.com/recurly/recurly-client-net/blob/f8ca4960b4b70570bddd559f70d3117f8f1bfc6d/Library/Tier.cs#L81-L83).

To handle the case where the user may not explicitly set `TierType` to `flat` when creating an add-on, we configure get and set accessors to set the default value for `TierType` to "flat". When creating a QBP add-on, `TierType` is required and will throw an error when one is not provided.

### DIFF
--- a/Library/AddOn.cs
+++ b/Library/AddOn.cs
@@ -29,7 +29,16 @@ namespace Recurly
         public DateTime UpdatedAt { get; private set; }
         public Adjustment.RevenueSchedule? RevenueScheduleType { get; set; }
         public string ItemState { get; set; }
-        public string TierType { get; set; }
+        private string _tierType;
+        public string TierType { 
+          get
+          {
+            return (_tierType == null) ? "flat" : _tierType;
+          }
+          set {
+            _tierType = value;
+          }
+        }
 
         private List<Tier> _tiers; 
 
@@ -221,7 +230,7 @@ namespace Recurly
         internal override void WriteXml(XmlTextWriter xmlWriter)
         {
             xmlWriter.WriteStartElement("add_on");
-            
+
             xmlWriter.WriteStringIfValid("item_code", ItemCode);
             xmlWriter.WriteStringIfValid("add_on_code", AddOnCode);
             xmlWriter.WriteStringIfValid("name", Name);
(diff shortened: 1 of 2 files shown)


Library/AddOn.cs
```diff
@@ -219,9 +219,7 @@ internal override void ReadXml(XmlTextReader reader)
         }
 
         internal override void WriteXml(XmlTextWriter xmlWriter)
-        {
-            xmlWriter.WriteStartElement("add_on");
```
bhelx, line 223: Why is this gone now?

bhelx: Nm, i see

Library/AddOn.cs
```diff
@@ -29,7 +29,18 @@ public enum Type
         public DateTime UpdatedAt { get; private set; }
         public Adjustment.RevenueSchedule? RevenueScheduleType { get; set; }
         public string ItemState { get; set; }
-        public string TierType { get; set; }
+        private string _tierType;
+        public string TierType { 
+          get
+          {
+            if (_tierType == null)
```
bhelx, line 36: I don't think you need to set the _tierType in the getter.

```
get
{
if (_tierType == null) return "flat";
return _tierType;
}
```

bhelx: Or maybe just:

```
get
{
  return (_tierType == null) ? "flat" : _tierType;
} 
```

joannasese: Much more eloquent!

PR #516, opened 2020-05-29
Ignore errors in build script

bhelx:
```
Installing dotnet formatter...
Tool 'dotnet-format' is already installed.
```

set -e it seems if dotnet is installed already, that command exits with
a bad status. This kills the whole script (keeping it from running build)

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-set -e
 
 if ! test -f "./bin/dotnet-format"; then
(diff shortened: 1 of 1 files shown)


scripts/build
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-set -e
```
bhelx, line 2: Maybe, instead i should figure out how to swallow the error with the dotnet tool command.

PR #518, opened 2020-05-29
Latest Features

bhelx:
* Adds `Client#record_external_transaction` operation
* Adds `PaymentMethod#name_on_account` attribute
* Adds `ShippingMethod#accounting_code` attribute
* Adds `SubscriptionAddOn#revenue_schedule_type` attribute
* Adds `state` query param to `Client#list_sites` operation


Recurly/Resources/Transaction.cs
```diff
@@ -81,7 +81,7 @@ public class Transaction : Resource
 
         /// <value>The values in this field will vary from gateway to gateway.</value>
         [JsonProperty("gateway_response_values")]
-        public Dictionary<string, string> GatewayResponseValues { get; set; }
+        public Dictionary<string, object> GatewayResponseValues { get; set; }
```
bhelx, line 84: Why is this here, this doesn't reflect what's in master?

PR #524, opened 2020-06-09
Adding RequestOptions to facilitate passing custom request headers

douglasmiller:
Defining the `RequestOptions` class and adding it as an optional parameter to each of the client operations. This will make it possible to pass custom headers (`AddHeader(key, value)`) to be included with individual requests.

The `RequestOptions` class currently provides the `AcceptLanguage` property with a setter and getter to facilitate setting the `Accept-Language` header. This is a shorthand version of `AddHeader("Accept-Language", value)`.

Additional properties will likely be added in the future as necessary.

Example usage:
```dotnet
var accountReq = new AccountCreate()
{
    // ...
};
var options = new RequestOptions {
    AcceptLanguage = "fr"
};
Account account = client.CreateAccount(accountReq, options);
```

### DIFF
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -98,6 +98,10 @@ namespace Recurly
 
         private RestRequest BuildRequest(Method method, string url, Request body = null, Dictionary<string, object> queryParams = null, RequestOptions options = null)
         {
+            if (options == null)
+            {
+                options = new RequestOptions();
+            }
             // If we have any query params, add them to the request
             if (queryParams != null)
             {
@@ -106,11 +110,7 @@ namespace Recurly
 
             var request = new RestRequest(url, method);
             request.JsonSerializer = Recurly.JsonSerializer.Default;
-
-            if (options != null)
-            {
-                request.AddHeaders(options.Headers);
-            }
+            request.AddHeaders(options.Headers);
 
             // If we have a body, serialize it and add it to the request
             if (body != null)
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -42,15 +42,16 @@ namespace Recurly
             );
             RestClient.AddHandler("application/json", () => { return new JsonSerializer(); });
 
+
             // These are the default headers to send on every request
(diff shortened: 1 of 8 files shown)


Recurly/BaseClient.cs
```diff
@@ -106,6 +107,11 @@ private RestRequest BuildRequest(Method method, string url, Request body = null,
             var request = new RestRequest(url, method);
             request.JsonSerializer = Recurly.JsonSerializer.Default;
 
+            if (options != null)
+            {
+                request.AddHeaders(options.Headers);
```
bhelx, line 112: How does this work? I'm assuming it preserves the current headers or else it wouldn't work. Does it override existing headers?

Recurly/BaseClient.cs
```diff
@@ -106,6 +107,11 @@ private RestRequest BuildRequest(Method method, string url, Request body = null,
             var request = new RestRequest(url, method);
             request.JsonSerializer = Recurly.JsonSerializer.Default;
 
+            if (options != null)
```
bhelx, line 110: Wondering if we shouldn't just initialize a default request options object instead then we don't need to do null checks everywhere. I could go either way.

```
if (options == null)
  options = new RequestOptions();
```

Recurly/BaseClient.cs
douglasmiller: It's a [convenience method on RestClient](https://github.com/restsharp/RestSharp/blob/master/src/RestSharp/RestRequest.cs#L396) that basically just calls `AddHeader` for each item in the collection. Yes, it does preserve existing headers.

Recurly/BaseClient.cs
bhelx: Or if csharp supports it:

```
, RequestOptions options = new RequestOptions())
```

douglasmiller: I went with your original suggestion.

Setting a default value in the method declaration would have to be done in the definition of the operation method. Still the `MakeRequest(Async)` and `BuildRequest` methods declare `options` as a nullable optional parameter. The `options` parameter could be moved before the other optional parameters in those methods, but I'm not sure if it's worth it.

PR #525, opened 2020-06-10
Add timeframe query params to cancel subscription endpoint

joannasese:
Adds the timeframe query params for the [cancel subscription endpoint](https://dev.recurly.com/docs/cancel-subscription).

### DIFF
--- a/Library/Subscription.cs
+++ b/Library/Subscription.cs
@@ -439,13 +439,14 @@ namespace Recurly
         /// </summary>
         public void Cancel(string timeframe = null)
         {
+          var url = UrlPrefix + Uri.EscapeDataString(Uuid) + "/cancel";
           if (timeframe == null) {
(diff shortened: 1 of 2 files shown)


Library/Subscription.cs
```diff
@@ -437,11 +437,17 @@ public void Create()
         /// Cancel an active subscription.  The subscription will not renew, but will continue to be active
         /// through the remainder of the current term.
         /// </summary>
-        public void Cancel()
+        public void Cancel(string timeframe = null)
         {
+          if (timeframe == null) {
```
bhelx, line 442: Maybe something like:

```csharp
var url = UrlPrefix + Uri.EscapeDataString(Uuid) + "/cancel";
if (timeframe == null)
   url += "?timeframe=" + timeframe";
```

joannasese: That's much better!

PR #528, opened 2020-06-17
BACS support

joannasese:
Added `sort_code` and `type` fields to support BACS payment type.

### DIFF
--- a/Test/BillingInfoTest.cs
+++ b/Test/BillingInfoTest.cs
@@ -109,6 +109,31 @@ namespace Recurly.Test
             }
             account2.BillingInfo.NameOnAccount.Should().Be("Iban account name");
         }
+
+        [RecurlyFact(TestEnvironment.Type.Integration)]
+        public void CreateBillingInfoWithBacs()
+        {
+            var account = CreateNewAccount();
+            var bacsInfo = new BillingInfo(account)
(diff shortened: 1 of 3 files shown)


Test/BillingInfoTest.cs
```diff
@@ -109,6 +109,31 @@ public void CreateBillingInfoWithIban()
...
+                Type = "bacs",
+            };
+            var threw = false;
+            try {
+                bacsInfo.Create();
+            } catch (ValidationException exception) {
+                threw = true;
+                exception.Errors[0].Symbol.Should().Be("card_type_not_accepted");
+            }
+            threw.Should().Be(true);
```
bhelx, line 135: Why does the only test for this intentionally fail? the error message seems confusing too

joannasese: Similar to our test with IBAN, to my knowledge we do not have a test account number for BACS the way we do for [credit cards or ACH](https://docs.recurly.com/docs/test). Therefore we'd expect this to fail. I borrowed this approach from the [`UpdateBillingInfoWithToken`](https://github.com/recurly/recurly-client-dotnet/blob/v2/Test/BillingInfoTest.cs#L68-L88) test. Is my logic flawed?

PR #539, opened 2020-06-24
Support items on subscriptions

joannasese:
This PR adds support for items on subscriptions, where merchants may now associate a subscription add-on with an item in their catalog.

`allow_any_item_on_subscriptions` has been added to the `Plan` class. It is used to determine whether items can be assigned as add-ons to individual subscriptions. If `true`, items can be assigned as add-ons to individual subscription add-ons. If `false`, only plan add-ons can be used.

`add_on_source`, added to the `SubscriptionAddOn` class, is used to determine where the associated add-on data is pulled from. If this value is set to `plan_add_on` or left blank, then add-on data will be pulled from the plan's add-ons. If the associated  `plan` has `allow_any_item_on_subscriptions` set to `true` and this field is set to `item`, then the associated add-on data will be pulled from the site's item catalog.

Code example:
```c#
var subscription = new Subscription(account, plan, "USD"); // account, plan, currency

// create sub add-on from item
subscription.AddOns.Add(item.ItemCode, "item", 299, 1); // add_on_code or item_code, add_on_source, unit_amount_in_cents, quantity

// create sub add-on from existing add-on
subscription.AddOns.Add(addon1.AddOnCode); 
// or
subscription.AddOns.Add(addon1.AddOnCode, "plan_add_on"); // add_on_code, add_on_source

subscription.Create();
```

### DIFF
--- a/Library/List/SubscriptionAddOnList.cs
+++ b/Library/List/SubscriptionAddOnList.cs
@@ -164,5 +164,20 @@ namespace Recurly
             var sub = new SubscriptionAddOn(planAddOnCode, addOnType, unitAmountInCents, quantity);
             base.Add(sub);
         }
+
+        // Supports item code as add-on code. Also supports plan add-on code as add-on code.
+        public void Add(string addOnCode, string addOnSource, int unitAmountInCents, int quantity = 1)
+        {
+          var sub = new SubscriptionAddOn(addOnCode, addOnSource, unitAmountInCents, quantity);
+          base.Add(sub);
+        }
+
+        public void Add(string planAddOnCode, string addOnSource, int quantity = 1)
+        {
+          var unitAmount = _subscription.Plan.AddOns.Find(ao => ao.AddOnCode == planAddOnCode).UnitAmountInCents[_subscription.Currency];
+          var sub = new SubscriptionAddOn(planAddOnCode, addOnSource, unitAmount, quantity);
+          base.Add(sub);
+        }
+
     }
 }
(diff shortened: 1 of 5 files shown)


Library/List/SubscriptionAddOnList.cs
```diff
@@ -164,5 +164,20 @@ public void Add(string planAddOnCode, AddOn.Type addOnType, int quantity, int un
...
+          var sub = new SubscriptionAddOn(addOnCode, addOnSource, unitAmountInCents, quantity);
+          base.Add(sub);
+        }
+
+        public void Add(string planAddOnCode, string addOnSource, int quantity = 1)
+        {
+          var unitAmount = _subscription.Plan.AddOns.Find(ao => ao.AddOnCode == planAddOnCode).UnitAmountInCents[_subscription.Currency];
+          var sub = new SubscriptionAddOn(planAddOnCode, addOnSource, unitAmount, quantity);
+          base.Add(sub);
+        }
```
joannasese, line 180: Not sure if this is the best approach, but it was an attempt to be comprehensive and cover different scenarios.

The first `Add()` covers the the crux of items on subscriptions functionality. 

- It accepts an item code for the first param. 
- Second param `addOnSource` is required and must be "item". 
- Third param `unitAmountInCents` is also required when `addOnSource = item`. 

However, this function should also support plan add-on codes. The user may want to pass in "plan_add_on" for `addOnSource` and may want to change the `unitAmountInCents`. This function allows them to do that.

The second `Add()` allows the user to pass in a value for `addOnSource` while inheriting the `unitAmountInCents` from the plan add-on.

PR #549, opened 2020-07-06
Adding Response class to Resource

douglasmiller:
The Response class is used to expose information about the response from Recurly API to the developer.
- The `Headers` property contains a List containing all of the response headers as Parameter objects.
- The `StatusCode` property contains the `HttpStatusCode` for the response.
- The `RawResponse` property contains the raw json response as a string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,20 @@ catch (Recurly.Errors.NetworkError ex)
 }
 ```
 
+### HTTP Metadata
+
+Sometimes you might want to get some additional information about the underlying HTTP request and response. Instead of returning this information directly and forcing the programmer to unwrap it, we inject this metadata into the top level resource that was returned. You can access the response by calling `GetResponse()` on any Resource.
+
+```csharp
+Account account = client.GetAccount(accountId);
+Response response = account.GetResponse();
+response.RawResponse // String body of the API response
+response.StatusCode // HTTP status code of the API response
+response.RequestId // "5b7019241a21d314-ATL"
+response.Headers // IList<Parameter> of all API response headers
+```
+
+
 ### Webhooks
 
 Recurly can send webhooks to any publicly accessible server.
--- a/README.md
+++ b/README.md
@@ -278,6 +278,12 @@ response.RequestId // "5b7019241a21d314-ATL"
 response.Headers // IList<Parameter> of all API response headers
 ```
 
+Rate Limit information is also accessible on the `Response` class. These values will be `null` when the corresponding headers are absent from the response. More information can be found on the developer portal's [Rate Limits](https://developers.recurly.com/api/v2019-10-10/index.html#section/Getting-Started/Limits) section.
+```csharp
+response.RateLimit // 2000  
+response.RateLimitRemaining // 1990
+response.RateLimitReset // 1595965380
+```
 
 ### Webhooks
 
--- a/Recurly.Tests/MockClient.cs
+++ b/Recurly.Tests/MockClient.cs
@@ -45,7 +45,7 @@ namespace Recurly.Tests
                     .Returns(item.Value.Object);
 
                 mockIRestClient
-                    .Setup(x => x.ExecuteTaskAsync<T>(It.Is<IRestRequest>(r => item.Key(r)), It.IsAny<CancellationToken>()))
+                    .Setup(x => x.ExecuteAsync<T>(It.Is<IRestRequest>(r => item.Key(r)), It.IsAny<CancellationToken>()))
                     .Returns(Task.FromResult(item.Value.Object));
             }
 
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -58,7 +58,7 @@ namespace Recurly
             {
                 var resp = t.Result;
                 this.HandleResponse(resp);
-                if (resp.Data is object)
+                if (resp.Data is Resource)
                     resp.Data.Response = Response.Build(resp);
                 return resp.Data;
             });
@@ -70,7 +70,7 @@ namespace Recurly
             var request = BuildRequest(method, url, body, queryParams, options);
             var resp = RestClient.Execute<T>(request);
             this.HandleResponse(resp);
-            if (resp.Data is object)
+            if (resp.Data is Resource)
                 resp.Data.Response = Response.Build(resp);
             return resp.Data;
         }
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Net;
 using System.Runtime.CompilerServices;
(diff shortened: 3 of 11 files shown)


Recurly/BaseClient.cs
```diff
@@ -48,25 +49,29 @@ public BaseClient(string apiKey)
             RestClient.AddDefaultHeader("Content-Type", "application/json");
         }
 
-        public async Task<T> MakeRequestAsync<T>(Method method, string url, Request body = null, Dictionary<string, object> queryParams = null, RequestOptions options = null, CancellationToken cancellationToken = default(CancellationToken)) where T : new()
+        public async Task<T> MakeRequestAsync<T>(Method method, string url, Request body = null, Dictionary<string, object> queryParams = null, RequestOptions options = null, CancellationToken cancellationToken = default(CancellationToken)) where T : Resource
         {
             Debug.WriteLine($"Calling {url}");
             var request = BuildRequest(method, url, body, queryParams, options);
-            var task = RestClient.ExecuteTaskAsync<T>(request, cancellationToken);
```
douglasmiller, line 55: Fixing deprecation warning.

Recurly/EmptyResource.cs
```diff
@@ -0,0 +1,7 @@
+using System.Diagnostics.CodeAnalysis;
+
+namespace Recurly
+{
+    [ExcludeFromCodeCoverage]
+    public class EmptyResource : Resource { }
```
douglasmiller, line 6: This is necessary to facilitate setting the base type of `MakeRequest` and `MakeRequestAsync` to `Response`.

This cannot be made available to the user without introducing a breaking change.

Recurly/BaseClient.cs
```diff
@@ -83,6 +88,7 @@ public int GetResourceCount(string url, Dictionary<string, object> queryParams)
             return int.Parse(recordCount);
         }
 
+        [ExcludeFromCodeCoverage]
```
cainj, line 91: Why are we excluding from Code Coverage?

bhelx: This `_` in the method name denotes it's not part of the public API. I could go either way on this, but it's technically not something that we need coverage for.

Recurly.Tests/MockClient.cs
```diff
@@ -45,7 +45,7 @@ internal static MockClient Build<T>(Dictionary<Func<IRestRequest, bool>, Mock<IR
                     .Returns(item.Value.Object);
 
                 mockIRestClient
-                    .Setup(x => x.ExecuteTaskAsync<T>(It.Is<IRestRequest>(r => item.Key(r)), It.IsAny<CancellationToken>()))
+                    .Setup(x => x.ExecuteAsync<T>(It.Is<IRestRequest>(r => item.Key(r)), It.IsAny<CancellationToken>()))
```
bhelx, line 48: Is this to fix the deprecation warning? The are equivalent behavior?

Recurly/Errors/ApiErrors.cs
```diff
@@ -128,96 +128,112 @@ public class BadRequest : ApiError
         public BadRequest(string message) : base(message) { }
         public BadRequest(string message, Exception inner) : base(message, inner) { }
     }
+    [ExcludeFromCodeCoverage]
```
bhelx, line 131: @cainj we generally also add these tags on generated code as this isn't really the responsibility of the library. Depends on how much "code" is actually there.

Recurly/BaseClient.cs
douglasmiller: `[SECURITY WARNING] _SetApiUrl is for testing only and not supported in production.`

Primarily because this method is 100% for internal smoke testing purposes.

Recurly/BaseClient.cs
```diff
@@ -48,25 +49,29 @@ public BaseClient(string apiKey)
...
         {
             Debug.WriteLine($"Calling {url}");
             var request = BuildRequest(method, url, body, queryParams, options);
-            var task = RestClient.ExecuteTaskAsync<T>(request, cancellationToken);
+            var task = RestClient.ExecuteAsync<T>(request, cancellationToken);
             return await task.ContinueWith(t =>
             {
                 var resp = t.Result;
                 this.HandleResponse(resp);
+                if (resp.Data is object)
```
bhelx, line 61: Shouldn't this be more specific? like `is Resource` or something?

Recurly/BaseClient.cs
cainj: sounds good.

Recurly.Tests/MockClient.cs
douglasmiller: Yes, this is to fix the deprecation warning and has the same behavior

Recurly/Errors/ApiErrors.cs
douglasmiller: Yes, this is actually restoring the `ExcludeFromCodeCoverage` that was mistakenly removed in a previous change in the generator.

Recurly/BaseClient.cs
douglasmiller: Updated, though `resp.Data` can only be a `Resource`

Recurly/Resource.cs
```diff
@@ -5,6 +5,7 @@ namespace Recurly
     [ExcludeFromCodeCoverage]
     public class Resource
     {
+        public Response Response { get; set; }
```
bhelx, line 8: Couple things to note:

1. Since we are exposing this, we need to document it. Let's add a section to the README similar to the docs we have for this feature in other languages
2. I'm okay with the `Response` property but this means that we can no longer support the key name `response` in json and we'd need to add this to the recurly-api-spec checker. My compromise for the other languages, while not idomatic, was to use a getter naming scheme `getResponse()`, `get_response`, `GetResponse()` etc.

douglasmiller: Updated README and adjusted the property

Recurly/Response.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public HttpStatusCode StatusCode { get; set; }
+
+        public IList<Parameter> Headers { get; set; }
+
+        public string RequestId { get { return GetHeader("X-Request-Id"); } }
+
+        public int? RateLimit { get { return GetIntHeader("X-RateLimit-Limit"); } }
+
+        public int? RateLimitRemaining { get { return GetIntHeader("X-RateLimit-Remaining"); } }
```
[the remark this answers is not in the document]
bhelx: Yes, it's similar to the `Option` type in scala. It wraps the primitive type in an object which can be inspected for `nullness`.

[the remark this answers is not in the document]
bhelx: I think it makes sense for `RecordCount`, but I believe these 3 rate limit headers are always present. Is there a case where they are not?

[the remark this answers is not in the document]
douglasmiller: I didn't encounter any scenarios in testing. I made them nullable in the event that an error is raised in a CDN layer or other location between the request and Recurly API.

README.md
```diff
@@ -265,6 +265,20 @@ catch (Recurly.Errors.NetworkError ex)
...
+Sometimes you might want to get some additional information about the underlying HTTP request and response. Instead of returning this information directly and forcing the programmer to unwrap it, we inject this metadata into the top level resource that was returned. You can access the response by calling `GetResponse()` on any Resource.
+
+```csharp
+Account account = client.GetAccount(accountId);
+Response response = account.GetResponse();
+response.RawResponse // String body of the API response
+response.StatusCode // HTTP status code of the API response
+response.RequestId // "5b7019241a21d314-ATL"
+response.Headers // IList<Parameter> of all API response headers
+```
```
bhelx, line 279: One more thing, can we add some examples of the rate limiting info as well?

Recurly/Resource.cs
```diff
@@ -5,6 +5,17 @@ namespace Recurly
...
     public class Resource
     {
+        private Response _response;
+
+        public Response GetResponse()
+        {
+            return _response;
+        }
+
+        public void SetResponse(Response response)
```
bhelx, line 15: I wonder if this should be `internal` rather than `public`. Perhaps we want to let people set it for testing purposes?

README.md
```diff
@@ -265,6 +265,26 @@ catch (Recurly.Errors.NetworkError ex)
...
+
+Rate Limit information is also accessible on the `Response` class. These values will be `null` when the corresponding headers are absent from the response. More information can be found on the developer portal's [Rate Limits](https://developers.recurly.com/api/v2019-10-10/index.html#section/Getting-Started/Limits) section.
+```csharp
```
bhelx, line 282: 👍

Recurly/Resource.cs
bhelx: Discussing offline.

Recurly/Response.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Recurly
+{
+    public class Response
+    {
+        public string RawResponse { get; set; }
+
+        public HttpStatusCode StatusCode { get; set; }
+
+        public IList<Parameter> Headers { get; set; }
```
bhelx: We decided here that we didn't want to expose `Parameter` which is from an internal dependency `RestSharp`

PR #554, opened 2020-07-28
Adds IClient interface and makes Client implement it

### LINKED ISSUE #527: Adding an interface for the Client. V3
There are a couple of nice advantages to having an interface exposed for the client and then have the client implement the interface. 
1. Allows for consumers to Mock the client for easy unit testing.
2. Allows for Dependency Injection.

bhelx:
Resolves #527 

The goal here was to simply supply the `IClient` interface and have the `Client` implement it. But not force the programmer to use it anywhere. It's just available if the programmer wants to use it in testing or DI.

### DIFF
--- a/Recurly/Errors/ApiErrors.cs
+++ b/Recurly/Errors/ApiErrors.cs
@@ -5,8 +5,8 @@
  * need and we will usher them to the appropriate places.
  */
 using System;
-using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


flyingtiger71:
@bhelx Thank you for this.  I have no additional comments.

bhelx:
@flyingtiger71 Great, opening this up for @douglasmiller to review

Issue #534, opened 2020-06-23
Ratelimit throttling Interface


flyingtiger71:
Provide Rate Limit Throttling Interface 

1. Provide an IRateLimitClient to be implemented by the consumer
2. In the base client MakeRequest and MakeRequestAsync methods add in a check to see if the call is a get/head call and if the IRateLimitClient has been implement.  If so then make a call to the RateLimitWaitExecute.
3. In the BaseClient HandleResponse if the RateLimitClient has been implemented then make the call to the IRateLimitClient SetRateLimitMetaData method passing the Limit, Remaining and Reset values from the X-RateLimie headers.

On the Client and BaseClient provide the constructors that would allow the consumer to inject their implementation of the IRateLimitClient.

```
namespace Recurly
{
    public interface IRateLimitClient
    {
        void RateLimitWaitExecute();
        void SetRateLimitMetaData(int rateLimit, int remaining, int fullReset);
    }
}
```
Example implementation
```
public class RateLimitClient : Recurly.IRateLimitClient
    {
        private string cacheKey;
        private ICacheClient cacheClient { get; set; }

        public RateLimitClient(string cacheKey, ICacheClient cacheClient)
        {
            this.cacheKey = cacheKey;
            this.cacheClient = cacheClient;
        }
        
        public void RateLimitWaitExecute()
        {
            if (cacheClient != null)
            {
                var rateLimitValue = cacheClient.Get<RateLimitValues>(cacheKey);
                if (rateLimitValue != null)
                {
                    var timeToSleepInMilliseconds = 0;
                    var percentage = rateLimitValue.RemaingCalls * 100.0 / rateLimitValue.Limit;
                    if(percentage < 5)
                    {
                        //Sleep for 10 seconds
                        timeToSleepInMilliseconds = 10000;
                    }
                    else if(percentage < 10)
                    {
                        //sleep for 5 seconds
                        timeToSleepInMilliseconds = 5000;
                    }
                    else if(percentage < 15)
                    {
                        //sleep for 2.5 seconds
                        timeToSleepInMilliseconds = 2500;
                    }
                    else if(percentage < 20)
                    {
                        //sleep for 1 seconds
                        timeToSleepInMilliseconds = 1000;
                    }
                    System.Threading.Thread.Sleep(timeToSleepInMilliseconds);
                }
            }
        }

        public void SetRateLimitMetaData(int rateLimit, int remaining, int fullReset)
        {
            if (cacheClient != null)
            {
[...]

PR #560, opened 2020-08-03
Client Event Handlers

bhelx:
Addresses #534 

Client EventHandlers are a generic interface for "hooking" into the http request / response events (and potentially other events the client might emit in the future). We'll be leaving it undocumented for the time being until we can get some more feedback. I've also left out async support for the time being. The generic handler interface can be used to do simple logging, or do something more interactive like a rate limit throttling.

EventHandler interface:

```csharp
namespace Recurly
{
    public interface IEventHandler
    {
        void OnRequest(Recurly.Http.Request request);
        void OnResponse(Recurly.Http.Response response);
    }
}
```
 Here is an example of a ratelimiter adapted from #534

```csharp
    class RateLimiter : Recurly.IEventHandler
    {
        private int? sleepTime;

        public RateLimiter() {
            this.sleepTime = null;
        }

        public void OnRequest(Recurly.Http.Request req)
        {
            if (this.sleepTime.HasValue)
            {
                Console.WriteLine($"Sleeping for {this.sleepTime / 1000} seconds");
                System.Threading.Thread.Sleep(this.sleepTime.Value);
            }
            else
            {
                Console.WriteLine("Not sleeping");
            }
        }

        public void OnResponse(Recurly.Http.Response resp)
        {
            if (resp.RateLimitRemaining.HasValue && resp.RateLimit.HasValue)
            {
                var percentage = resp.RateLimitRemaining.Value * 100.0 / resp.RateLimit.Value;
                Console.WriteLine($"Percentage: {percentage}");
                if (percentage < 5)
                {
                    //Sleep for 10 seconds
                    this.sleepTime = 10000;
                }
                else if (percentage < 10)
                {
                    //sleep for 5 seconds
                    this.sleepTime = 5000;
                }
                else if (percentage < 15)
                {
                    //sleep for 2.5 seconds
                    this.sleepTime = 2500;
                }
                else if (percentage < 20)
                {
                    //sleep for 1 seconds
                    this.sleepTime = 1000;
                }
                else
                {
                    this.sleepTime = null;
                }
            }
        }
    }
```

It can be registered with:

```csharp
client.AddEventHandler(new RateLimiter());
```

### DIFF
--- a/scripts/format
+++ b/scripts/format
@@ -2,7 +2,7 @@
 set -e
 
 if [ "$1" == "--check" ]; then
-  ~/.dotnet/tools/dotnet-format --check --dry-run
+  ~/.dotnet/tools/dotnet-format --check
 else
   ~/.dotnet/tools/dotnet-format
 fi
--- a/Recurly.Tests/BaseClientTest.cs
+++ b/Recurly.Tests/BaseClientTest.cs
@@ -124,6 +124,20 @@ namespace Recurly.Tests
             Assert.Throws<Recurly.RecurlyError>(() => client.GetResource("benjamin", "param1", new DateTime(2020, 01, 01)));
         }
 
+        [Fact]
+        public void WillTriggerHookIfAvailable()
+        {
+            var client = MockClient.Build(SuccessResponse(System.Net.HttpStatusCode.OK));
+            var mockHandler = new Mock<IEventHandler>();
+            mockHandler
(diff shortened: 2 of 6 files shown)


scripts/format
```diff
@@ -2,7 +2,7 @@
 set -e
 
 if [ "$1" == "--check" ]; then
-  ~/.dotnet/tools/dotnet-format --check --dry-run
```
bhelx, line 5: this option was deprecated for --check

PR #561, opened 2020-08-06
To enable updating QBP sub add-ons, add Tiers to SubscriptionAddOn

joannasese:
Note that this change will trigger the following error because `unit_amount_in_cents` is sent as a dictionary of currencies and values instead of as an `Integer`. This error will be addressed in a separate ticket. `The provided XML was invalid. | Field: "" Code: "" Symbol: "invalid_xml" Details: "Tag <unit_amount_in_cents> must contain only text"`

Sample code:
```c#
var subscription = Subscriptions.Get(subscription.Uuid);
var change = new SubscriptionChange {
  TimeFrame = SubscriptionChange.ChangeTimeframe.Now,
  AddOns = new SubscriptionAddOnList()
};

SubscriptionAddOn subaddon = subscription.AddOns[0];

var tier1 = subaddon.Tiers[0];
tier1.UnitAmountInCents.Add("USD", 99);
tier1.EndingQuantity = 11;

change.AddOns.Add(subaddon); 

Subscription.ChangeSubscription(subscription.Uuid, change);
```

### DIFF
--- a/Library/SubscriptionAddOn.cs
+++ b/Library/SubscriptionAddOn.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Xml;
 using System.Globalization;
+using System.Collections.Generic;
 
 namespace Recurly
 {
@@ -13,6 +14,18 @@ namespace Recurly
         public Adjustment.RevenueSchedule? RevenueScheduleType { get; set; }
         public float? UsagePercentage { get; set; }
         public string TierType { get; set; }
+
+        private List<Tier> _tiers; 
+
+        /// <summary>
+        /// List of tiers for this add-on
+        /// </summary>
+        public List<Tier> Tiers
+        {
+            get {return _tiers ?? (_tiers = new List<Tier>()); }
+            set { _tiers = value; }
+        }
+
         public string AddOnSource { get; set; }
 
         public SubscriptionAddOn(XmlTextReader reader)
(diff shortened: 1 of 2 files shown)


Library/SubscriptionAddOn.cs
```diff
@@ -100,6 +113,20 @@ internal override void ReadXml(XmlTextReader reader)
...
                         break;
 
+                    case "tiers":
+                        Tiers = new List<Tier>();
+                        while (reader.Read())
+                        {
+                            if (reader.Name == "tiers" && reader.NodeType == XmlNodeType.EndElement)
+                                break;
+
+                            if (reader.NodeType == XmlNodeType.Element && reader.Name == "tier")
```
cainj, line 123: Nitpicky:  It makes for cleaner code to use `else if` if possible to clarify that the branches are exclusive.

cainj:
Overall looking good.

Library/SubscriptionAddOn.cs
joannasese: Good call!

joannasese:
@cainj I did not includes tests for this (yet!) because attempts to update a QBP subscription add-on will fail until the "money bug" is repaired. Once we address the currency vs integer issue, this functionality will work as expected and then I'll add tests.

PR #564, opened 2020-08-13
Apply level of abstraction to xml writers by using WriteStringIfValid()

joannasese:
Cleaned up some xml writers.

It makes sense to replace
```c#
if (!AccountNumber.IsNullOrEmpty())
{
  xmlWriter.WriteElementString("account_number", AccountNumber);
}
```
with `xmlWriter.WriteStringIfValid("account_number", AccountNumber);` because

```c#
public static void WriteStringIfValid(this XmlWriter writer, string localName, string value)
{
  if (!value.IsNullOrEmpty())
    writer.WriteElementString(localName, value);
}
```

### DIFF
--- a/Library/BillingInfo.cs
+++ b/Library/BillingInfo.cs
@@ -430,6 +430,14 @@ namespace Recurly
                 xmlWriter.WriteStringIfValid("phone", PhoneNumber);
                 xmlWriter.WriteStringIfValid("vat_number", VatNumber);
                 xmlWriter.WriteStringIfValid("currency", Currency);
+                xmlWriter.WriteStringIfValid("account_number", AccountNumber);
+                xmlWriter.WriteStringIfValid("type", Type);
+                xmlWriter.WriteStringIfValid("sort_code", SortCode);
+                xmlWriter.WriteStringIfValid("bsb_code", BsbCode);
+                xmlWriter.WriteStringIfValid("iban", Iban);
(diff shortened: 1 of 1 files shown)


Library/BillingInfo.cs
```diff
@@ -430,6 +430,15 @@ internal override void WriteXml(XmlTextWriter xmlWriter)
...
                 xmlWriter.WriteStringIfValid("currency", Currency);
+                xmlWriter.WriteStringIfValid("account_number", AccountNumber);
+                xmlWriter.WriteStringIfValid("type", Type);
+                xmlWriter.WriteStringIfValid("sort_code", SortCode);
+                xmlWriter.WriteStringIfValid("bsb_code", BsbCode);
+                xmlWriter.WriteStringIfValid("iban", Iban);
+                xmlWriter.WriteStringIfValid("paypal_billing_agreement_id", PaypalBillingAgreementId);
+                xmlWriter.WriteStringIfValid("amazon_billing_agreement_id", AmazonBillingAgreementId);
+                xmlWriter.WriteStringIfValid("amazon_region", AmazonRegion);
+                xmlWriter.WriteStringIfValid("external_hpp_type", ExternalHppType.Value.ToString().EnumNameToTransportCase());
```
bhelx, line 441: This call should probably stay the same. by calling `.Value` you are assuming it's always gonna be there. it's unsafe to operate on the value without first checking if it's null or not.

PR #566, opened 2020-08-17
Write xml for address2 field as long as it is not null

joannasese:
Reported issue:
The Address2 field in the BillingInfo object cannot be cleared by sending an empty string ("").
```c#
var account = Accounts.Get("account_code");
var info = new BillingInfo(account);
info.Address2 = ""; // Does not clear Address2
info.Update();
```

### DIFF
--- a/Library/BillingInfo.cs
+++ b/Library/BillingInfo.cs
@@ -422,7 +422,7 @@ namespace Recurly
                 xmlWriter.WriteStringIfValid("company", Company);
                 xmlWriter.WriteStringIfValid("name_on_account", NameOnAccount);
                 xmlWriter.WriteStringIfValid("address1", Address1);
-                xmlWriter.WriteStringIfValid("address2", Address2);
+                if (Address2 != null) xmlWriter.WriteElementString("address2", Address2);
                 xmlWriter.WriteStringIfValid("city", City);
                 xmlWriter.WriteStringIfValid("state", State);
                 xmlWriter.WriteStringIfValid("zip", PostalCode);


Library/BillingInfo.cs
```diff
@@ -422,7 +422,7 @@ internal override void WriteXml(XmlTextWriter xmlWriter)
                 xmlWriter.WriteStringIfValid("company", Company);
                 xmlWriter.WriteStringIfValid("name_on_account", NameOnAccount);
                 xmlWriter.WriteStringIfValid("address1", Address1);
-                xmlWriter.WriteStringIfValid("address2", Address2);
+                if (Address2 != null) xmlWriter.WriteElementString("address2", Address2);
```
joannasese, line 425: Would it be more correct as two lines?

cainj: I don't think so.  I was trying to find the `WriteStringIfValid` but I think the function does something similar to what you are doing with these two lines

joannasese: `WriteStringIfValid` is defined here: https://github.com/recurly/recurly-client-dotnet/blob/v2/Library/Extensions/XmlWriterExtensions.cs#L16-L20

@bhelx had suggested an `if not null` approach, for which there is precedent here: https://github.com/recurly/recurly-client-dotnet/blob/v2/Library/Address.cs#L84-L112

Arguably we could apply this to all the address fields, since the fields are not required and the user should be given an option to clear out an existing address without replacing it. Just not sure how far we want to go.

cainj: I think we should keep it consistent.  I like the `extension` function approach, seems a little cleaner.  I this case, let's use the `WriteStringIfValid` to keep it consistent with the code in this file.

joannasese: The issue with using `WriteStringIfValid()` is that it includes this logic:
```c#
if (!value.IsNullOrEmpty())
                writer.WriteElementString(localName, value);
```

However, if we want to be able to clear out a field using `""`, we can't utilize `IsNullOrEmpty()`. Which is too bad, cause it's looks cleaner that way!

cainj: I guess I don't quite understand the problem that's being solved.  We can either have a bunch of `if/else` logic in this file or let the extension function `WriteStringIfValid()` encapsulate `if/else` for us? 
If I were a new developer to this code I think it's easier to look using `WriteStringIfValid()` then seeing a bunch of `if/else` statements.

PR #570, opened 2020-08-26
Adding hierarchical errors and status code error generation

douglasmiller:
Non-breaking update to error classes to give them a hierarchical structure. This will facilitate differentiating between a `Error.Client` error and a `Errors.Server` error for example.

New error classes have also been added to be parent classes for some existing errors (e.g. `Errors.Forbidden` error is a new class and both `Errors.InvalidApiKey` error and `Errors.InvalidPermissions` error extend it).

Most importantly, this update allows the client library to gracefully handle non-json errors in the rare occasions when they occur. These errors will be based on the HTTP status code of the response and will fall back to the existing ApiError if there is not a defined mapping.

This is the `4.x` version of https://github.com/recurly/recurly-client-dotnet/pull/569

### DIFF
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -154,7 +154,17 @@ namespace Recurly
             // has likely occurred
             if (resp.ErrorException != null)
             {
-                throw new RecurlyError(resp.ErrorMessage);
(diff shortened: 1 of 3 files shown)


douglasmiller:
Since this is a duplicate of #569, we will just merge master into the v3-v2020-01-01 branch to implement the updates

PR #577, opened 2020-09-15
New endpoint to verify an account's billing information

joannasese:
If the account has billing information for anything other than a credit card, this will return 422, with an appropriate error message. If successful, this will return 200, with the transaction's data.
`gateway_code` (optional) is the code for the gateway to use for the verification. If unspecified, a gateway will be selected using the normal rules.

Example
```c#
// verify without specifying gateway code to use 
billingInfo.Verify()
// verify with a specific gateway code
billingInfo.Verify("gateway-code")
```

### DIFF
--- a/Library/VerifyGateway.cs
+++ b/Library/VerifyGateway.cs
@@ -1,0 +1,38 @@
+using System;
+using System.Xml;
+
+namespace Recurly
+{
+    class VerifyGateway : RecurlyEntity
+    {
+        public string GatewayCode { get; set; }
+
+        internal VerifyGateway(string gateway_code)
+        {
+            GatewayCode = gateway_code;
+        }
+
+        internal override void ReadXml(XmlTextReader reader)
+            {
+            while (reader.Read())
+            {
+              if (reader.Name == "verify" && reader.NodeType == XmlNodeType.EndElement)
+                  break;
+
+              if (reader.NodeType != XmlNodeType.Element) continue;
+
+              if (reader.Name == "gateway_code")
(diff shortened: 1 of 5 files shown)


Library/VerifyGateway.cs
```diff
@@ -0,0 +1,38 @@
+using System;
+using System.Xml;
+
+namespace Recurly
+{
+    class VerifyGateway : RecurlyEntity
```
douglasmiller, line 6: Would `VerifyBillingInfo` be a more accurate name?

Library/VerifyGateway.cs
```diff
@@ -0,0 +1,38 @@
...
+              if (reader.Name == "gateway_code")
+                    GatewayCode = reader.ReadElementContentAsString();
+            }
+        }
+
+        internal override void WriteXml(XmlTextWriter writer)
+        {
+            writer.WriteStartElement("verify");
+
+            writer.WriteElementString("gateway_code", GatewayCode);
```
douglasmiller, line 33: I think that if you only include this if `GatewayCode` is not null/empty, then the conditional checks in the `Verify` method can be removed entirely.

Library/VerifyGateway.cs
joannasese: probably.

Library/VerifyGateway.cs
joannasese: Following up on discussion, you'll see some adjustments. Looks much better.

douglasmiller (approved):
👍

PR #582, opened 2020-10-07
Adding OptionalParams and per-operation implementations

douglasmiller:
This is an update for the `4.x` version of the client.

Adds the `OptionalParams` class and implementations for each operation that accepts optional parameters. This is to enable the addition of new optional parameters in a manner that will not introduce potential breaking changes as experienced in https://github.com/recurly/recurly-client-dotnet/pull/568. 

3.x Implementation
```csharp
var accounts = client.ListAccounts(limit: 200);
```

4.x Implementation
```csharp
var optionalParams = new ListAccountsParams()
{
    Limit = 200
};
var accounts = client.ListAccounts(optionalParams);
```


joannasese (approved):
👍

PR #584, opened 2020-10-13
Adding support for List types in optional query params

douglasmiller:
This is an update for the `4.x` version of the client and is an extension of https://github.com/recurly/recurly-client-dotnet/pull/582.

This update will allow a developer to specify List/Array optional parameters as a List instead of needing to generate a comma separated string.

3.x Implementation
```csharp
var accounts = client.ListAccounts(ids: "account-id-1,account-id-2,account-id-3");
```

4.x Implementation
```csharp
var optionalParams = new ListAccountsParams()
{
    Ids = new List<string>() {
      "account-id-1",
      "account-id-2",
      "account-id-3",
    }
};
var accounts = client.ListAccounts(optionalParams);
```

### DIFF
--- a/Recurly.Tests/Fixtures.cs
+++ b/Recurly.Tests/Fixtures.cs
@@ -57,7 +57,7 @@ namespace Recurly.Tests
     public class ListResourcesParams : OptionalParams
     {
         [JsonProperty("ids")]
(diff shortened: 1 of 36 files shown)


joannasese (approved):
👍

PR #369, opened 2019-01-09
Add amazon_region to Billing Info

aaron-suarez:
This adds the new amazon_region attribute to the Billing Info class.

```C#
namespace TestRig
{
    class CreateBillingInfo
    {
        public static void Run(string[] args)
        {
            var account = Accounts.Get("x");
            var info = new BillingInfo(account);
            info.FirstName = "Aaron";
            info.LastName = "Du Monde";
            info.Address1 = "123 Main St";
            info.City = "New Orleans";
            info.State = "LA";
            info.Country = "US";
            info.PostalCode = "70212";
            info.AmazonBillingAgreementId = "C01-1234567-8901234";
            info.AmazonRegion = "eu";

            try
            {
                info.Create();
            }
            catch(RecurlyException e)
            {
                Console.WriteLine(e.Errors[0]);
            }
        }
    }
}
```

PR #587, opened 2020-10-27
Add test coverage for unknown error types

jguidry-recurly:
Add a test case to verify that unknown error types are thrown as `ApiError`. I'm testing for the desired behavior [by way of `ArgumentException`](https://github.com/recurly/recurly-client-dotnet/blob/master/Recurly/Errors/Factory.cs#L189) which feels wonky. I'm open to any suggestions for a better way to work around [strict mode](https://github.com/recurly/recurly-client-dotnet/blob/master/Recurly/Utils.cs#L11) at test time to allow a more direct test.

See related [#640 in recurly-client-ruby](https://github.com/recurly/recurly-client-ruby/issues/640) and [#369 in recurly-client-python](https://github.com/recurly/recurly-client-python/issues/369) for context.

### DIFF
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -195,7 +195,7 @@ namespace Recurly
             // has likely occurred
             if (resp.ErrorException != null)
             {
-                var message = !resp.ErrorException.Equals(null) ? resp.ErrorMessage : $"Unexpected {resp.StatusCode} Error.";
+                var message = resp.ErrorMessage;
                 if (resp.Headers.Any(t => t.Name == "X-Request-ID"))
                 {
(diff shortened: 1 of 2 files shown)


joannasese (approved):
This LGTM but I'd want another opinion on it before merging, perhaps from noted C# developer @douglasmiller

Recurly/BaseClient.cs
```diff
@@ -195,7 +195,7 @@ private void HandleResponse(IRestResponse resp)
             // has likely occurred
             if (resp.ErrorException != null)
             {
-                var message = !resp.ErrorException.Equals(null) ? resp.ErrorMessage : $"Unexpected {resp.StatusCode} Error.";
```
douglasmiller, line 198: 👍

douglasmiller (approved):
👍

PR #588, opened 2020-11-03
Support item-specific coupons

joannasese:
Add `item_codes` and `applies_to_all_items` to Coupon

Example
```c#
var coupon = new Coupon(
  "special",
  "Special", 
  new Dictionary<string, int>() { { "USD", 200 } }
);
// `applies_to_all_items` defaults to false, so next line is optional
coupon.AppliesToAllItems = false;
coupon.Items.Add("one_item");
coupon.Items.Add("two_item");
coupon.Create();

```

### DIFF
--- a/Test/CouponTest.cs
+++ b/Test/CouponTest.cs
@@ -160,6 +160,25 @@ namespace Recurly.Test
         }
 
         [RecurlyFact(TestEnvironment.Type.Integration)]
+        public void CreateCouponItem()
+        {
+            var item = new Item(GetMockItemCode("coupon item"), "Coupon Item");
+            item.Create();
+
+            var anotherItem = new Item(GetMockItemCode("another item"), "Another Item");
+            anotherItem.Create();
+
+            var coupon = new Coupon(GetMockCouponCode(), GetMockCouponName(), new Dictionary<string, int>());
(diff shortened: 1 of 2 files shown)


Test/CouponTest.cs
```diff
@@ -159,6 +159,25 @@ public void Coupon_plan_must_exist()
...
+            var item = new Item(GetMockItemCode("coupon item"), "Coupon Item");
+            item.Create();
+
+            var anotherItem = new Item(GetMockItemCode("another item"), "Another Item");
+            anotherItem.Create();
+
+            var coupon = new Coupon(GetMockCouponCode(), GetMockCouponName(), new Dictionary<string, int>());
+            coupon.DiscountInCents.Add("USD", 200);
+            coupon.Items.Add(item.ItemCode);
+            coupon.Items.Add(anotherItem.ItemCode);
```
jguidry-recurly, line 174: Since it is being added, is it worth flexing `AppliesToAllItems` here? It seems like a basic value passthrough, so not sure how much mileage it would net.

joannasese: Good call. I think it's worth adding the coverage.

jguidry-recurly (approved):
LGTM

douglasmiller (approved):
:+1:

PR #593, opened 2020-11-13
Expose fields in response xml to developer

joannasese:
- Because `item_state` and `external_sku` are included in the xml response for item-based add-ons, it would be good practice to expose this fields to the programmer.
- Same applies to `last_two` for `BillingInfo`


douglasmiller (approved):
👍

PR #594, opened 2020-11-17
Latest Changes for 2019-10-10

douglasmiller:
- Adds the `state` parameter to the `ListAccountCouponRedemptions` operation.
- Adds the `charge` parameter to the `TerminateSubscription` operation so that a merchant can choose to not bill a customer for un-billed usage when terminating a subscription


joannasese (approved):
👍

PR #601, opened 2020-12-21
Configure bump2version

joannasese:
Add ability to bump client library version with a script instead of needing to do so manually


joannasese:
Closing, as this gets folded up in https://github.com/recurly/recurly-client-dotnet/pull/629

PR #602, opened 2021-01-15
Repair Lookup Credit Payment endpoint

joannasese:
- Correct request url from /credit_payment/ to /credit_payments/
- Fix AmountInCents attribute, which was previously included as UnitAmountInCent
- Added test for credit payments

Previous calls to [GET /credit_payments/{uuid}](https://developers.recurly.com/api-v2/v2.29/index.html#operation/lookupCreditPayment) would fail due to incorrect request url. Simply adding an 's' resolves the issue

### DIFF
--- a/Library/CreditPayment.cs
+++ b/Library/CreditPayment.cs
@@ -10,7 +10,7 @@ namespace Recurly
     {
         public string Uuid { get; set; }
         public string Action { get; set; }
-        public int UnitAmountInCents { get; set; }
+        public int AmountInCents { get; set; }
         public string AppliedToInvoice { get; set; }
         public string Currency { get; set; }
         public DateTime CreatedAt { get; set; }
@@ -49,8 +49,8 @@ namespace Recurly
                         Uuid = reader.ReadElementContentAsString();
                         break;
         
-                    case "unit_amount_in_cents":
-                        UnitAmountInCents = reader.ReadElementContentAsInt();
+                    case "amount_in_cents":
+                        AmountInCents = reader.ReadElementContentAsInt();
                         break;
    
                     case "currency":
@@ -109,7 +109,7 @@ namespace Recurly
 
             var creditPayment = new CreditPayment();
             Client.Instance.PerformRequest(Client.HttpRequestMethod.Get,
-                "/credit_payment/" + Uri.EscapeDataString(uuid),
+                "/credit_payments/" + Uri.EscapeDataString(uuid),
                 creditPayment.ReadXml);
             return creditPayment;
         }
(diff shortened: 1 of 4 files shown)


Test/CreditPaymentTest.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Recurly.Test
+{
+  public class CreditPaymentTest : BaseTest
+  {
+    [RecurlyFact(TestEnvironment.Type.Integration)]
+    public void LookupCreditPayment()
+    {
+      var collection = CreateNewCollection();
+      var transactionUuid = collection.ChargeInvoice.Transactions[0].Uuid;
+      var cancelledCollection = Purchase.Cancel(transactionUuid);
```
jguidry-recurly, line 15: For my own learnings: why does the purchase need to be canceled? Pre-emptive cleanup?

Library/CreditPayment.cs
```diff
@@ -10,7 +10,7 @@ public class CreditPayment : RecurlyEntity
     {
         public string Uuid { get; set; }
         public string Action { get; set; }
-        public int UnitAmountInCents { get; set; }
+        public int AmountInCents { get; set; }
```
jguidry-recurly, line 13: This isn't a breaking change as the old attribute was incorrect, right?

jguidry-recurly (approved):
👍

Test/CreditPaymentTest.cs
joannasese: Canceling the purchase authorization triggers a credit payment on the account. Here, we're authorizing a purchase ([it's abstracted under `CreateNewCollection`](https://github.com/recurly/recurly-client-dotnet/blob/v2/Test/BaseTest.cs#L76-L90)), grabbing the transaction uuid, using that uuid to cancel the purchase authorization, then finally fetching the credit payment that was created when we canceled.

Library/CreditPayment.cs
joannasese: Hmmm. Well the old attribute was incorrect in that it would always return a value of 0. I'm not sure if this counts as a bug fix or a breaking change, in that yes, someone's code will break if they try to parse `UnitAmountInCents` (they will encounter `error CS1061: 'CreditPayment' does not contain a definition for 'UnitAmountInCents'`). If we treat this as a breaking change, I think I ought to change the base branch for this PR.

jguidry-recurly: > code will break if they try to parse

Maybe I have a misunderstanding. Is this strictly for parsing or did it expose a public interface i.e. `creditPayment.UnitAmountInCents`?

PR #608, opened 2021-01-25
Expose HTTP timeout for Recurly.BaseClient

jguidry-recurly:
Given that the unit tests mock the HTTP client at the RestSharp interface level, I was unable to automate validation of the timeout functionality. A simple test has been added to verify that the property can be initialized and read back.

I did however modify some playground scripts and manually confirm that it was behaving as expected.

image

In the past, I've verified configured timeouts by making a real HTTP connection to a debug server launched by the test runner. Unclear how to approach that in this repo as:

1. overriding the endpoint URL at test time is tricky
2. unclear if a barebones Kestrel server can be raised at test time

Open to suggestions for improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ var client = new Recurly.Client(apiKey);
 var sub = client.GetSubscription("uuid-abcd123456")
 ```
 
+Optional arguments can be provided through object initializers.
+
+```csharp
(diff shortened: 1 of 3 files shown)


jguidry-recurly:
Another thought: is there an obvious place to document this? I think this is the first client we've done.

joannasese (approved):
LGTM

PR #610, opened 2021-02-03
usageId type needs to match nullable type

joannasese:
`usageId` type needs to match nullable type established in Id getter/setter in order for `Get` to function properly (for example, when passing in a variable as a param instead of the actual id itself).

This is valid:
```c#
Usages.Get("374ae5e848adcfd332fdd3469d89c888", "video_streaming", 450646065398417338);
```

This fails with error `error CS1503: Argument 3: cannot convert from 'long?' to 'long'`
```c#
Usages.Get("374ae5e848adcfd332fdd3469d89c888", "video_streaming", usage.Id)
```

Will need to add testing for Usage, since it doesn't currently exist, but that will be in the future

### DIFF
--- a/Library/Usage.cs
+++ b/Library/Usage.cs
@@ -272,7 +272,7 @@ namespace Recurly
             return new UsageList(UrlPrefix(subscriptionUuid, subscriptionAddOnCode) + "?" + parameters.ToString());
         }
 
-        public static Usage Get(string subscriptionUuid, string subscriptionAddOnCode, long usageId)
+        public static Usage Get(string subscriptionUuid, string subscriptionAddOnCode, long? usageId)
         {
             if (string.IsNullOrWhiteSpace(subscriptionUuid) || string.IsNullOrWhiteSpace(subscriptionAddOnCode))
             {


Library/Usage.cs
```diff
@@ -272,7 +272,7 @@ public static RecurlyList<Usage> List(string subscriptionUuid, string subscripti
             return new UsageList(UrlPrefix(subscriptionUuid, subscriptionAddOnCode) + "?" + parameters.ToString());
         }
 
-        public static Usage Get(string subscriptionUuid, string subscriptionAddOnCode, long usageId)
+        public static Usage Get(string subscriptionUuid, string subscriptionAddOnCode, long? usageId)
```
jguidry-recurly, line 275: Now that this can be `null`, does it need to add guard logic or a fallback? L285 looks like the only place `usageId` is referenced. It's been a while, but unlike say, Ruby, I don't think `null` has a `ToString` method in C#. Either way though -- runtime error or constructing a URL with the literal string`"null"` -- feels like it might be an issue.

joannasese: I'm putting a pause on this for now since the audience affected is minimal and I'd like to do more research on `Id` here: https://github.com/recurly/recurly-client-dotnet/blob/v2/Library/Usage.cs#L19.

PR #612, opened 2021-02-19
Updating changelog script and changelog generator config for 4.x release

douglasmiller:


scripts/changelog
```diff
@@ -7,7 +7,7 @@ else
   if [ "$1" == "--pending" ]; then
     github_changelog_generator -t $GITHUB_TOKEN --unreleased-only -o "$2"
   elif [ "$1" == "--release-tag" ]; then
-    github_changelog_generator -t $GITHUB_TOKEN --unreleased true --unreleased-label "$2"
+    github_changelog_generator -t $GITHUB_TOKEN --unreleased true --future-release "$2"
```
douglasmiller, line 10: This should have always been `--future-release` instead of `--unreleased-label`. I only recently discovered the difference.

scripts/build
```diff
@@ -3,7 +3,7 @@ set -e
 
 if ! command -v dotnet-format; then
   echo "Installing dotnet formatter..."
-  dotnet tool install -g dotnet-format
+  dotnet tool install -g dotnet-format --version 4.0.0
```
douglasmiller, line 6: I had to lock this down since the 5.x version causes this error:
```
It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '2.1.0-rc1' was not found.
  - Check application dependencies and target a framework version installed at:
      /usr/local/share/dotnet
  - Installing .NET Core prerequisites might help resolve this problem:
      https://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
  - The .NET Core framework and SDK can be installed from:
      https://aka.ms/dotnet-download
  - The following versions are installed:
      2.2.8 at [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]

```

jguidry-recurly: Not sure -- I encountered some similar errors when I brought the dotnet integration example app into the latest Docker image. I agree with limiting the version until we can look into the broader issue.

jguidry-recurly (approved):

PR #639, opened 2021-06-03
Postpone subscriptions using the field next_bill_date

cyberxander90:
Pospone subscriptions using the `next_bill_date ` field instead of the deprecated `next_renewal_date`.
[DOC](https://developers.recurly.com/api-v2/v2.29/index.html#operation/postponeSubscriptionOrExtendTrial)

### DIFF
--- a/Library/Subscription.cs
+++ b/Library/Subscription.cs
@@ -496,7 +496,10 @@ namespace Recurly
         /// <summary>
         /// For an active subscription, this will pause the subscription until the specified date.
         /// </summary>
-        /// <param name="nextRenewalDate">The specified time the subscription will be postponed</param>
+        /// <param name="nextRenewalDate">The specified time the subscription will be postponed.
+        /// Although next_renewal_date is deprecated and has been replaced with next_bill_date,
+        /// we have opted to maintain the param name nextRenewalDate to keep compatibility with named arguments.
+        /// </param>
         /// <param name="bulk">bulk = false (default) or true to bypass the 60 second wait while postponing</param>
         public void Postpone(DateTime nextRenewalDate, bool bulk = false)
         {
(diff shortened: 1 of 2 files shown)


Library/Subscription.cs
```diff
@@ -496,14 +496,14 @@ public void Preview()
         /// <summary>
         /// For an active subscription, this will pause the subscription until the specified date.
         /// </summary>
-        /// <param name="nextRenewalDate">The specified time the subscription will be postponed</param>
+        /// <param name="nextBillDate">The specified time the subscription will be postponed</param>
         /// <param name="bulk">bulk = false (default) or true to bypass the 60 second wait while postponing</param>
-        public void Postpone(DateTime nextRenewalDate, bool bulk = false)
```
douglasmiller, line 501: I'm concerned about this being a breaking change for integrations that are using [Named Arguments](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments), as `subscription.Postpone(nextRenewalDate: new DateTime(2022, 1, 1))` would no longer work.

I'd like to get input from @joannasese about this.

cyberxander90: You are right.
I think we could keep the parameter name as `nextRenewalDate` and only update the URL query string.
What do you think?

joannasese: Nice callout. @cyberxander90 I guess the only thing I would add is to augment the `<summary>` with a quick explanation explaining why `<param name="nextRenewalDate">` remains, citing named arguments.

Test/SubscriptionTest.cs
```diff
@@ -539,7 +539,7 @@ public void UpdateNotesSubscription()
         {
             var plan = new Plan(GetMockPlanCode(), GetMockPlanName())
             {
-                Description = "Postpone Subscription Test"
+                Description = "Update Notes Subscription Test"
```
joannasese, line 542: Nice catch! Appreciate your cleanup here.

joannasese (approved):
👍

PR #658, opened 2021-09-17
Fixing bug in the Pager that breaks the Count and First methods

douglasmiller:
When the `FetchNextPage` or `FetchNextPageAsync` methods were called on the `Pager`, the private method `Clone` would copy properties from the result set into the instance of the pager. The method was unintentionally nulling out properties of the pager instance that would not be present on the result set.

This bug fix will only clone properties that will be present on the result set.


jguidry-recurly (approved):
LGTM

PR #661, opened 2021-10-12
Add bill_for_account to Adjustment

FabricioCoutinho:


Test/AdjustmentTest.cs
```diff
@@ -148,7 +148,7 @@ public void ListAdjustmentsPendingToInvoiced()
             account.InvoicePendingCharges();
 
             adjustments = account.GetAdjustments(state: Adjustment.AdjustmentState.Invoiced);
-            adjustments.Should().HaveCount(2);
+            adjustments.Should().HaveCount(3);
```
icas-recurly: I am curious, why does the adjustment count changed?

icas-recurly (changes requested):

Test/AdjustmentTest.cs
FabricioCoutinho: isn't changing and the test is falling on local but passing on CI checks. I fixed and now is passing on local environment.

jguidry-recurly (approved):

PR #664, opened 2021-10-18
Client library update for V2 Dunning Campaign API enhancements

sunnyw1212:
Add dunning_campaign_id to Account, Invoice, and Plan
PUT v2/dunning_campaigns/{dunning_campaign_id}/bulk_update

### DIFF
--- a/Library/DunningCampaign.cs
+++ b/Library/DunningCampaign.cs
@@ -1,0 +1,190 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Xml;
+
+namespace Recurly
+{
+
+  /// <summary>
+  /// A Dunning Campaign in Recurly.
+  /// </summary>
+  public class DunningCampaign : RecurlyEntity
+  {
+    /// <summary>
+    /// Unique ID assigned to this dunning campaign.
+    /// </summary>
+    public string Id { get; private set; }
+
+    public string Name { get; private set; }
+
+    public string Code { get; private set; }
+
+    public string Description { get; private set; }
+
+    public bool DefaultCampaign { get; private set; }
+
+    private List<DunningCycle> _dunningCycles;
+
+    public List<DunningCycle> DunningCycles
+    {
+        get { return _dunningCycles ?? (_dunningCycles = new List<DunningCycle>()); }
+        set { _dunningCycles = value; }
+    }
+
+    public DateTime? CreatedAt { get; private set; }
+
+    public DateTime? UpdatedAt { get; private set; }
+
+    public DateTime? DeletedAt { get; private set; }
+
+    internal const string UrlPrefix = "/dunning_campaigns/";
+
+    private List<string> PlanCodes { get; set;}
+
+    #region Constructors
+
+    public DunningCampaign()
+    {
+    }
+
+    internal DunningCampaign(XmlTextReader xmlReader)
+    {
+      ReadXml(xmlReader);
+    }
+
+    #endregion
+
+    /// <summary>
+    /// Update the Dunning Campaign for multiple Plans
+    /// </summary>
+    public void BulkUpdate(List<string> planCodes)
+    {
+        PlanCodes = new List<string>();
+        foreach (string planCode in planCodes)
+        {
+          PlanCodes.Add(planCode);
+        }
+        // PUT /dunning_campaigns/<id>/bulk_update
+        Client.Instance.PerformRequest(Client.HttpRequestMethod.Put,
+            UrlPrefix + Uri.EscapeDataString(Id.ToString()) + "/bulk_update",
+            WriteBulkUpdateXml);
+    }
+
+    internal override void ReadXml(XmlTextReader reader)
+    {
+      while (reader.Read())
+      {
+        if (reader.Name == "dunning_campaign" && reader.NodeType == XmlNodeType.EndElement)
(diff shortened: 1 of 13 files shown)


Library/Interval.cs
```diff
@@ -0,0 +1,50 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Xml;
+
+namespace Recurly
+{
+
+  public class Interval : RecurlyEntity
```
jguidry-recurly: Do you think we should call it `DunningInterval` to keep these classes/files grouped? Given the reference to email template, it's a bit more specific than a generic interval class. Minor nit.

judith (approved):

Library/DunningCampaign.cs
```diff
@@ -0,0 +1,189 @@
...
+
+    public DateTime? CreatedAt { get; private set; }
+
+    public DateTime? UpdatedAt { get; private set; }
+
+    public DateTime? DeletedAt { get; private set; }
+
+    internal const string UrlPrefix = "/dunning_campaigns/";
+
+    public List<string> PlanCodes { get; private set;}
```
douglasmiller: I think that we want to make this private as there is no reason to expose the availability of this variable to the developer. It is used solely as a storage location between `BulkUpdate` being called and the `WriteBulkUpdateXML` method being executed.

```suggestion
    private List<string> PlanCodes { get; private set;}
```

Library/DunningCampaign.cs
```diff
@@ -0,0 +1,189 @@
...
+            DunningCycles.Add(new DunningCycle(reader));
+        }
+      }
+    }
+
+    internal void WriteBulkUpdateXml(XmlTextWriter xmlWriter)
+    {
+      xmlWriter.WriteStartElement("dunning_campaign");
+      xmlWriter.WriteStartElement("plan_codes");
+      foreach (string planCode in PlanCodes) {
```
douglasmiller: Nit: C# puts opening curly braces on a new line

```suggestion
      foreach (string planCode in PlanCodes)
      {
```

Library/List/DunningCampaignList.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Collections.Specialized;
+using System.Net;
+using System.Web;
+using System.Xml;
+
+namespace Recurly
+{
+    public class DunningCampaignList : RecurlyList<DunningCampaign>
```
douglasmiller, line 9: Did we determine that the API will return a pager shaped response even though we have a limit on the total number of dunning campaigns?

douglasmiller (changes requested):

Library/Interval.cs
sunnyw1212: hmm its's a good point, but we're calling it `Interval` in all the other client libs and passing it back as `interval` in the xml response. 

not sure if it's worth going back and changing that name everywhere at this point?

Library/DunningCampaign.cs
sunnyw1212: yep good call - fixed

Library/DunningCampaign.cs
sunnyw1212: fixed

Library/Interval.cs
jguidry-recurly: On further reflection, agree. Remaining consistent (within reason) across client implementations is a good priority and, as you point out, the naming is already blessed in the underlying schema.

I actually steered Amanda toward `DunningInterval` [on the Python client yesterday](https://github.com/recurly/recurly-client-python/pull/515/files#diff-29341c2217f213f31f173ac977f56d2eba7834da42c33b76ed85fb8673e82920R525) though. @amandamfielding Do you have any leaning on this? I'm thinking that it should made to match.

Library/List/DunningCampaignList.cs
sunnyw1212: yes, as far as i can tell we are

douglasmiller (approved):
👍🏼 

Deferring to @jguidry-recurly for final approval.

Library/Interval.cs
amandamfielding: @jguidry-recurly I think we should stay consistent and i'll make the change in the python library to call it `Interval`

judith: Actually after discussing it further we decided that it makes sense to call the resource for the `interval` in the XML response `DunningInterval` in the client libraries as it will match what is happening in the API V3/V4 client libraries.

Library/DunningCycle.cs
```diff
@@ -0,0 +1,120 @@
...
+                break;
+
+            if (reader.NodeType != XmlNodeType.Element) continue;
+
+            switch (reader.Name)
+            {
+                case "type":
+                    Type = reader.ReadElementContentAsString();
+                    break;
+                case "applies_to_manual":
```
jguidry-recurly: Should this be `applies_to_manual_trial`? I encountered [a problem](https://github.com/recurly/recurly-client-python/pull/515#discussion_r733037553) with this on the Python client.

sunnyw1212: good catch - fixed

judith (approved):

jguidry-recurly (approved):

douglasmiller (approved):

PR #679, opened 2022-01-19
Add region argument to client to connect in EU data center

FabricioCoutinho:
Add an ability to define which Data Center the client will connect

### DIFF
--- a/Recurly/BaseClient.cs
+++ b/Recurly/BaseClient.cs
@@ -18,21 +18,38 @@ namespace Recurly
     public class BaseClient
     {
         private string ApiKey { get; }
-        private const string ApiUrl = "https://v3.recurly.com/";
+        private Dictionary<string, string> ApiHosts = new Dictionary<string, string>()
+        {
+            { "us", "https://v3.recurly.com/" },
+            { "eu", "https://v3.eu.recurly.com/" }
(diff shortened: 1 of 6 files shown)


Recurly/BaseClient.cs
```diff
@@ -19,20 +19,21 @@ public class BaseClient
...
         private const string ApiUrl = "https://v3.recurly.com/";
+        private const string EuApiUrl = "https://v3.eu.recurly.com/";
         private string[] BinaryTypes = { "application/pdf" };
         private List<IEventHandler> EventHandlers = new List<IEventHandler>();
         public virtual string ApiVersion { get; protected set; }
 
         internal IRestClient RestClient { get; set; }
 
-        public BaseClient(string apiKey)
+        public BaseClient(string apiKey, string region = "us")
```
douglasmiller: Instead of having these hard coded strings (`"us"` and `"eu"`), lets add an enum for these:

```csharp
public enum Region
{
    US,
    EU
}
```

douglasmiller:
I have opened another PR (https://github.com/recurly/recurly-client-dotnet/pull/681) that introduces an alternative approach.

douglasmiller (approved):

PR #684, opened 2022-02-07
V2: Add invoice templates

icas-recurly:
* Add `InvoiceTemplate` resource
```
InvoiceTemplates.Get('q0tzf7o7fpbl')
InvoiceTemplates.List()
```

* Add `invoice_template_uuid ` and `invoice_template` to `Account`
```
var account = Accounts.Get("acc"');
account.InvoiceTemplateUuid = "q0tzf7o7fpbl";
account.Update();
account.GetInvoiceTemplate().Code;
```

### DIFF
--- a/Library/Account.cs
+++ b/Library/Account.cs
@@ -54,6 +54,7 @@ namespace Recurly
         public string ParentAccountCode { get; set; }
         public string TransactionType { get; set; }
         public string DunningCampaignId { get; set; }
+        public string InvoiceTemplateUuid { get; set; }
 
         private AccountAcquisition _accountAcquisition;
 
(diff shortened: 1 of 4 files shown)


Library/Account.cs
```diff
@@ -224,7 +225,8 @@ public void Update()
             // PUT /accounts/<account code>
             Client.Instance.PerformRequest(Client.HttpRequestMethod.Put,
                 UrlPrefix + Uri.EscapeDataString(AccountCode),
-                WriteXml);
+                WriteXml,
+                ReadXml);
```
icas-recurly, line 229: @naspthorp @chrissrogers in order to make this library work as the others when we perform the update(at least with account model) i had to pass the ReadXml so it overrides the the current state based on the latest info on the server. For example, if i update the invoice template uuid, the account response will include an href pointing to the new reference. 
I am honestly a little bit scare about doing this change

icas-recurly: Also, it is bad that Account will work different from the other classes

icas-recurly: Maybe i can create an special XMLReader that just update the invoice template _invoiceTemplateUuid private instance var

naspthorp (approved):

chrissrogers (approved):

PR #685, opened 2022-02-07
Add online_banking_payment_type to billing_info

cyberxander90:


Library/BillingInfo.cs
```diff
@@ -36,6 +36,12 @@ public enum BankAccountType : short
             Savings
         }
 
+        public enum OnlineBankingPmtType : short
+        {
+            Ideal,
+            Sofort
+        }
```
brittany-wood, line 43: aren't these supposed to be lower case?

PR #707, opened 2022-05-02
Add support for percentage tiers at plan and subscription le…

camilopineda100:


seckenrode (commented):
This looks good from my side, but I'm going to tag @btruncali1 in for a light review since he's got more professional C# experience

Library/CurrencyPercentageTier.cs
```diff
@@ -0,0 +1,71 @@
...
+    public List<PercentageTier> PercentageTiers
+    {
+        get { return _percentageTiers ?? (_percentageTiers = new List<PercentageTier>()); }
+        set { _percentageTiers = value; }
+    }
+
+    #region Constructors
+
+    public CurrencyPercentageTier() {
+    }
```
btruncali1: id do an empty constructor in 1 line, but up to you totally

Library/CurrencyPercentageTier.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    internal CurrencyPercentageTier(XmlTextReader xmlReader){
+      ReadXml(xmlReader);
```
btruncali1, line 28: ifd do this but up to you

`internal CurrencyPercentageTier(XmlTextReader xmlReader)
{

      ReadXml(xmlReader);

 }`

btruncali1: yikes its hard to get styling right in these comments

btruncali1 (approved):

PR #720, opened 2022-08-04
Added Support for Ramp Pricing Feature

btruncali1:
Added support for ramp pricing on plans.

Add `PlanRampInterval` and `PlanRampPricing` models, as well as `PricingModelType` enum.
Add `List<PlanRampInterval>` and `PricingModel` to the Plan object.
Added support for reading XML responses and writing XML requests for ramp pricing on a plan.
Added unit tests for plan `get`, `create`, and `update` for ramp priced plans.

### DIFF
--- a/Library/Plan.cs
+++ b/Library/Plan.cs
@@ -78,6 +78,22 @@ namespace Recurly
             }
         }
 
+        /// <summary>
+        /// The pricing model type for the plan.  Can be a 'fixed' price plan or a 'ramp' priced plan
+        /// </summary>
+        public PricingModelType? PricingModel { get; set; }
+
+        /// <summary>
+        /// The ramp intervals representing the pricing schedule for the subscription
+        /// </summary>
+        public List<PlanRampInterval> RampIntervals
+        {
+            get { return _rampIntervals ?? (_rampIntervals = new List<PlanRampInterval>()); }
+            set { _rampIntervals = value; }
+        }
+
+        private List<PlanRampInterval> _rampIntervals;
+
         private Dictionary<string, int> _unitAmountInCents;
         /// <summary>
         /// A dictionary of currencies and values for the subscription amount
@@ -233,10 +249,26 @@ namespace Recurly
             }
         }
 
+        internal void ReadXmlRampIntervals(XmlTextReader reader)
+        {
+            while (reader.Read())
+            {
+                if ((reader.Name == "ramp_intervals") &&
+                    reader.NodeType == XmlNodeType.EndElement)
+                    break;
+
+                if (reader.NodeType == XmlNodeType.Element && reader.Name == "ramp_interval")
+                {
+                    RampIntervals.Add(new PlanRampInterval(reader));
+                }
+            }
+        }
+
         internal override void ReadXml(XmlTextReader reader)
         {
             UnitAmountInCents.Clear();
             SetupFeeInCents.Clear();
+            RampIntervals.Clear();
 
             while (reader.Read())
             {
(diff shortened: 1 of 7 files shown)


Library/Plan.cs
```diff
@@ -449,6 +489,19 @@ internal override void WriteXml(XmlTextWriter xmlWriter)
             if (AutoRenew.HasValue)
                 xmlWriter.WriteElementString("auto_renew", AutoRenew.Value.AsString());
 
+            if (!PricingModel.ToString().IsNullOrEmpty())
+                xmlWriter.WriteElementString("pricing_model", PricingModel.ToString().EnumNameToTransportCase());
```
btruncali1, line 493: make sure that during an update if its not set it doesnt get sent

Library/PlanRampInterval.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Xml;
+
+namespace Recurly
+{
+    public class PlanRampInterval : RecurlyEntity
+    {
+        /// <summary>Represents the first billing cycle of a ramp.</summary>
+        public int? StartingBillingCycle { get; set; }
```
csampson: Should this be nullable?

Library/PlanRampPricing.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Recurly
+{
+    public class PlanRampPricing : RecurlyEntity
+    {
+        /// <summary>3-letter ISO 4217 currency code.</summary>
+        public string Currency { get; set; }
+
+        /// <summary>Represents the price for the Ramp Interval.</summary>
+        public int? UnitAmountInCents { get; set; }
```
csampson: Should be non-nullable, right?

Library/PlanRampPricing.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public PlanRampPricing() { }
```
csampson, line 14: Should we just omit the constructor? Or is the idea to leave it open to future updates?

Library/Plan.cs
```diff
@@ -449,6 +489,19 @@ internal override void WriteXml(XmlTextWriter xmlWriter)
             if (AutoRenew.HasValue)
                 xmlWriter.WriteElementString("auto_renew", AutoRenew.Value.AsString());
 
+            if (!PricingModel.ToString().IsNullOrEmpty())
```
csampson: Should `PricingModel` be nullable then? Seems like we could just use `HasValue` here, if it were.

Library/PlanRampPricing.cs
btruncali1: Yea i thought this may come up.  This is how it was generated for v3.  Im not sure if some decision was made to have them generate like this for safety and then the server does all the validation?  Both unitAmount and StartingBilling cycle were generated by v3 like that.  I dont have an issue removing the nullability tho

Library/PlanRampPricing.cs
btruncali1: This is required for instantiation.

Library/PlanRampPricing.cs
btruncali1: It _could_ have something to do with like, an integer if its not nullible, and you dont set it, in c# it defaults to 0.  And so if you dont set it, perhaps the thinking was, we would rather allow you to do that and send you a server side error, rather than having it pass in zero... honestly im not sure

Library/Plan.cs
btruncali1: umm yes, yes it should, good call

Library/PlanRampPricing.cs
btruncali1: All of that said, if its nullable, it looks like you dont have to set it, which i dont like... So it would read clearer to me if it was not nullable.

csampson (approved):

PR #724, opened 2022-08-16
Add Unit Tests for Purchases with Ramp Pricing

btruncali1:
Added unit tests to support Purchases with ramp priced subscriptions.

Added the following unit tests: `InvoicePurchaseWithRampPlan`, `InvoicePurchaseWithCustomRamps`, `PreviewPurchaseWithRamps`, `PendingPurchaseWithRamps`, `AuthAndCapturePurchaseWithRamps`

### DIFF
--- a/Test/BaseTest.cs
+++ b/Test/BaseTest.cs
@@ -268,6 +268,7 @@ namespace Recurly.Test
         public Plan CreateNewRampPlan(int numberOfRamps)
         {
             var plan = new Plan(GetMockPlanCode(), GetMockPlanName()) { Description = "Create Ramp Plan" };
+            plan.TaxExempt = true;
             plan.SetupFeeInCents.Add("USD", 0);
             plan.PricingModel = PricingModelType.Ramp;
             plan.RampIntervals = GetMockPlanRampIntervals(numberOfRamps);
--- a/Test/PurchaseTest.cs
+++ b/Test/PurchaseTest.cs
@@ -30,6 +30,112 @@ namespace Recurly.Test
         }
 
         [RecurlyFact(TestEnvironment.Type.Integration)]
+        public void InvoicePurchaseWithRampPlan()
+        {
+            var account = CreateNewAccountWithBillingInfo();
+            var plan = CreateNewRampPlan(3);
+            PlansToDeactivateOnDispose.Add(plan);
+
+            var purchase = new Purchase(account.AccountCode, "USD");
+            purchase.Account.BillingInfo = account.BillingInfo;
+
+            var sub = new Subscription(plan.PlanCode);
+            purchase.Subscriptions.Add(sub);
+
+            var collection = Purchase.Invoice(purchase);
+            int firstRampUnitAmount = plan.RampIntervals[0].Currencies[0].UnitAmountInCents;
+
+            Assert.NotNull(collection.ChargeInvoice);
+            Assert.Equal(collection.ChargeInvoice.SubtotalInCents, firstRampUnitAmount);
+            Assert.Equal(collection.ChargeInvoice.TotalInCents, firstRampUnitAmount);
+            Assert.Equal(collection.ChargeInvoice.SubtotalBeforeDiscountInCents, firstRampUnitAmount);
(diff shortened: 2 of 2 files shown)


Test/PurchaseTest.cs
```diff
@@ -29,6 +29,116 @@ public void InvoicePurchase()
             Assert.Equal(collection.ChargeInvoice.Address.Company, "Acme Software");
         }
 
+        [RecurlyFact(TestEnvironment.Type.Integration)]
+        public void InvoicePurchaseWithRampPlan()
+        {
+            var account = CreateNewAccountWithBillingInfo();
+            var currency = "USD";
```
csampson, line 36: nit: could just hardcode this

Test/PurchaseTest.cs
```diff
@@ -29,6 +29,116 @@ public void InvoicePurchase()
...
+            PlansToDeactivateOnDispose.Add(plan);
+
+            var purchase = new Purchase(account.AccountCode, currency);
+            purchase.Account.BillingInfo = account.BillingInfo;
+
+            var sub = new Subscription(plan.PlanCode);
+            purchase.Subscriptions.Add(sub);
+
+            var collection = Purchase.Invoice(purchase);
+            int firstRampUnitAmount = plan.RampIntervals[0].Currencies[0].UnitAmountInCents;
```
csampson, line 46: For my own understanding, shouldn't `var` be fine here, since the type is easily inferred?

Test/PurchaseTest.cs
csampson: nit: could just hardcode these currency values

Test/PurchaseTest.cs
```diff
@@ -29,6 +29,116 @@ public void InvoicePurchase()
...
+            PlansToDeactivateOnDispose.Add(plan);
+            var account = CreateNewAccountWithBillingInfo();
+
+            var sub = new Subscription(account, plan, "USD");
+            sub.RampIntervals = GetMockSubscriptionRampIntervals(2);
+
+            var purchase = new Purchase(sub.AccountCode, currency);
+            purchase.Account.BillingInfo = sub.Account.BillingInfo;
+            // Randomly generate an email address
+            purchase.Account.Email = Guid.NewGuid().ToString().Substring(0, 8) + "@test.com";
```
csampson, line 123: nit: any reason not to just "user@example.com"?

mwlang (approved):
LGTM.  Some back and forth on what's what in Slack with Brett, but overall, this is pretty straightforward and cleanly implemented.

Test/BaseTest.cs
```diff
@@ -268,6 +268,7 @@ public Subscription CreateNewRampSubscription(int numberOfRamps)
         public Plan CreateNewRampPlan(int numberOfRamps)
         {
             var plan = new Plan(GetMockPlanCode(), GetMockPlanName()) { Description = "Create Ramp Plan" };
+            plan.TaxExempt = true;
```
csampson, line 271: What's the reason for this change?

btruncali1: Because in the prod site, if i dont do this then taxes are included in the Totals coming back and therefore they dont match the first ramp value

btruncali1: Comes from this in the AC:

```
Associated plans have NO tax.
Associated purchases have NO gift cards or other discounts
```

Test/PurchaseTest.cs
btruncali1: Just to prevent the test from being flakey.  i guess i wasnt wanting accounts to be out there all with the same email incase something somehow didnt make it to the account.close().  Everything else in the tests are setup to get new random string, for example plan_code or plan name.

Test/PurchaseTest.cs
btruncali1: it could be "fine" there.  But better to be explicit when you can.

btruncali1: Typically when its the base types like `int` or `string` or `decimal`, youre explicit about it in c# when you can be

csampson (approved):

PR #735, opened 2022-10-10
adds support for entitlements

amandamfielding:
GET accounts/{account_id}/entitlements

### DIFF
--- a/Library/Entitlement.cs
+++ b/Library/Entitlement.cs
@@ -1,0 +1,103 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Xml;
+
+namespace Recurly
+{
+
+    public class Entitlement : RecurlyEntity
+    {
+        public string AccountCode { get; private set; }
+
+        public CustomerPermission CustomerPermission { get; private set; }
+
+        private List<string> _grantedBy;
+
+        public List<string> GrantedBy
+        {
+            get { return _grantedBy ?? (_grantedBy = new List<string>()); }
+        }
+
+        public DateTime? CreatedAt { get; private set; }
+
+        public DateTime? UpdatedAt { get; private set; }
+
+        internal Entitlement(string accountCode) : this()
+        {
+            AccountCode = accountCode;
+        }
+
+        public Entitlement(Account account) : this()
+        {
+            AccountCode = account.AccountCode;
+        }
+
+        internal Entitlement(XmlTextReader xmlReader)
+        {
+            ReadXml(xmlReader);
+        }
+
+        public Entitlement() { }
+
+        internal override void ReadXml(XmlTextReader reader)
+        {
+            while (reader.Read())
+            {
+                if (reader.Name == "entitlement" && reader.NodeType == XmlNodeType.EndElement)
+                    break;
+
+                if (reader.NodeType != XmlNodeType.Element) continue;
+
+                switch (reader.Name)
+                {
+                    case "entitlement":
+                        break;
+
+                    case "account":
+                        var href = reader.GetAttribute("href");
+                        AccountCode = Uri.UnescapeDataString(href.Substring(href.LastIndexOf("/") + 1));
+                        break;
+
+                    case "customer_permission":
+                        CustomerPermission = new CustomerPermission(reader);
+                        break;
+
+                    case "granted_by":
+                        ReadXMLGrantedBys(reader);
+                        break;
+
+                    case "created_at":
+                        CreatedAt = reader.ReadElementContentAsDateTime();
+                        break;
+
+                    case "updated_at":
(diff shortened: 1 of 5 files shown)


Library/GrantedBy.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Xml;
+
+namespace Recurly
+{
+    /// <summary>
```
btruncali1: This doesnt seem to be doing anything, what about taking your ReadXMLGrantedBys method above and removing it from Entitlement and implementing it in this class instead?   I think that would encapsulate responsibilities better.

btruncali1:
Are you planning on adding some unit tests to this PR?

Library/Entitlement.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        public Entitlement() { }
+
+        /// <summary>
+        /// Lookup a Recurly account's entitlements
+        /// </summary>
+        /// <param name="accountCode"></param>
+        /// <returns></returns>
+        public static Entitlement Get(string accountCode)
```
btruncali1: It doesnt look like this is ever getting called

Library/Entitlement.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Recurly
+{
+
+    public class Entitlement : RecurlyEntity
+    {
+        public string AccountCode { get; private set; }
+
+        public CustomerPermission CustomerPermission { get; private set; }
+
+        public List<GrantedBy> GrantedBys { get; private set; }
```
btruncali1: This would need to change to 
        `public GrantedBy GrantedBy { get; private set; }`

btruncali1: That would get rid of those Github Action errors further down

Library/GrantedBy.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// Subscriptions and Items granting a Customer Permission in Recurly.
+    ///
+    /// </summary>
+    public class GrantedBy : RecurlyEntity
+    {
```
btruncali1: Im actually not sure if this should be of type `SubscriptionList` or `List<Subscription>`, or if it matters, we will want to play with it seeing what the GetSubscriptions method returns.  Same with the ItemList

Library/Entitlement.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Collections.Generic;
+using System.Net;
+using System.Xml;
+
+namespace Recurly
+{
+
+    public class Entitlement : RecurlyEntity
+    {
+        public string AccountCode { get; private set; }
```
btruncali1, line 11: If Account code isnt supposed to be here lets remove it.  Otherwise we need to set it because its currently null

Library/Entitlement.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            var statusCode = Client.Instance.PerformRequest(Client.HttpRequestMethod.Get,
+                EntitlementUrl(accountCode),
+                entitlement.ReadXml);
+
+            return statusCode == HttpStatusCode.NotFound ? null : entitlement;
+        }
+
+        private static string EntitlementUrl(string accountCode)
+        {
```
btruncali1: also unused

Library/Entitlement.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public CustomerPermission CustomerPermission { get; private set; }
+
+        public GrantedBy GrantedBy { get; private set; }
+
+        public DateTime? CreatedAt { get; private set; }
+
+        public DateTime? UpdatedAt { get; private set; }
+
+        private const string UrlPrefix = "/accounts/";
```
btruncali1: these 2 url members are not used

Library/List/EntitlementList.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Specialized;
+using System.Net;
+using System.Web;
+using System.Xml;
+
+namespace Recurly
+{
+    public class EntitlementList : RecurlyList<Entitlement>
+    {
+        internal EntitlementList(string baseUrl) : base(Client.HttpRequestMethod.Get, baseUrl)
```
btruncali1, line 11: i dont think this needs to inherit from client, because like i mentioned in the comments above we are not doing any api calls from Entitlement.  The `GetEntitlement()` is happening in Account.cs

amandamfielding: I was following what we do on NoteList and BillingInfoList and without this I don't get my entitlements array

btruncali1: ya i see now, it is required

btruncali1 (approved):
LGTM!

PR #736, opened 2022-10-12
Entitlements test brett

btruncali1:


Library/Entitlement.cs
```diff
@@ -0,0 +1,130 @@
...
+                    break;
+
+                if (reader.NodeType != XmlNodeType.Element) continue;
+
+
+                if (reader.NodeType == XmlNodeType.Element)
+                {
+                    var href = reader.GetAttribute("href");
+                    string id = href.Substring(href.LastIndexOf("/") + 1);
+                    GrantedBy.Add(reader.Name, id);
```
btruncali1, line 119: @amandamfielding a Dictionary in c# need a unique keys, so if you had multiple subscriptions and "subscription" is the key like it is here, it would break.  We could solve that by making the `id` the key instead of `reader.Name` (just swap `reader.Name` and `id` in the above line), OR instead of using a dictionary we could use a generic `KeyValuePair<string, string>` type.

PR #750, opened 2022-11-09
Support for external subscriptions and products

ELepolt:
GET /external_subscriptions
GET /external_subscriptions/:id
GET /accounts/:account_code/external_subscriptions
GET /external_products
GET /external_products/:id

### DIFF
--- a/Library/ExternalProduct.cs
+++ b/Library/ExternalProduct.cs
@@ -1,0 +1,184 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Xml;
+
+namespace Recurly
+{
+
+    /// <summary>
+    /// Represents external_products for accounts
+    /// </summary>
+    public class ExternalProduct : RecurlyEntity
+    {
+        private Plan _plan;
+        public Plan Plan
+        {
+            get { return _plan ?? (_plan = Plans.Get(PlanCode)); }
+            set
+            {
+                _plan = value;
+                PlanCode = value.PlanCode;
+            }
+        }
+        public string PlanCode { get; private set; }
+        public string Name { get; private set; }
+        public string ExternalObjectReference { get; private set; }
+        public List<ExternalProductReference> ExternalProductReferenceList { get; private set; }
+        public DateTime CreatedAt { get; private set; }
+        public DateTime UpdatedAt { get; private set; }
+
+        internal const string UrlPrefix = "/external_products/";
+
+        internal ExternalProduct()
+        {
+        }
+
+        internal ExternalProduct(XmlTextReader reader)
+        {
+            ReadXml(reader);
+        }
+
+
+        #region Read and Write XML documents
+
+        internal void ReadExternalResourceXml(XmlTextReader reader)
+        {
+            while (reader.Read())
+            {
+                if (reader.Name == "external_resource" && reader.NodeType == XmlNodeType.EndElement)
+                    break;
+
+                if (reader.NodeType != XmlNodeType.Element) continue;
+
+                switch (reader.Name)
+                {
+                    case "external_object_reference":
+                        ExternalObjectReference = reader.ReadElementContentAsString();
+                        break;
+                }
+            }
+        }
+
+        internal void ReadExternalProductReferenceXml(XmlTextReader reader)
+        {
+            ExternalProductReferenceList = new List<ExternalProductReference>();
+            while (reader.Read())
+            {
+                if (reader.Name == "external_product_references" && reader.NodeType == XmlNodeType.EndElement)
+                    break;
+
+                if (reader.NodeType != XmlNodeType.Element) continue;
+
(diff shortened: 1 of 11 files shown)


Library/ExternalProduct.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// <summary>
+    /// Represents external_products for accounts
+    /// </summary>
+    public class ExternalProduct : RecurlyEntity
+    {
+
+
+
+        public string Name { get; private set; }
+        public string PlanName { get; private set; }
```
Patrick-Duvall: It looks like the top level non-nested nodes on an external_product are `name, created_at, updated_at` and the nested nodes are `plan` and `external_product_references`.  Is it an issue for this class that planName and PlanCode are directly on the `ExternalProduct`?

Library/ExternalSubscription.cs
```diff
@@ -0,0 +1,189 @@
...
+        public string AccountCode => _accountCode;
+
+        private Account _account;
+        public Account Account
+        {
+            get { return _account ?? (_account = Accounts.Get(_accountCode)); }
+            internal set { _account = value; }
+        }
+
+        public string ExternalObjectReference { get; private set; }
```
Patrick-Duvall: same question About this being nested under `external_resource`

Library/ExternalSubscription.cs
```diff
@@ -0,0 +1,189 @@
...
+            return new ExternalSubscriptionList(ExternalSubscription.UrlPrefix + "?" + parameters.ToString());
+        }
+
+        public static ExternalSubscription Get(string uuid)
+        {
+            if (string.IsNullOrWhiteSpace(uuid))
+            {
+                return null;
+            }
+            var s = new ExternalSubscription();
```
Patrick-Duvall: I personally don't love single letters as variable names, are you open to changing this `external_sub`

Test/ExternalSubscriptionTest.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Recurly.Test
+{
+    public class ExternalSubscriptionTest : BaseTest
+    {
+        [RecurlyFact(TestEnvironment.Type.Integration)]
+        public void LookupExternalSubscription()
+        {
+            var uuid = "rv1loyqlf6si";
+            var externalSubscription = ExternalSubscriptions.Get(uuid);
+            externalSubscription.Quantity.Should().Be(14);
```
Patrick-Duvall, line 15: Do you know why getting a thing by uuid is returning a collection of 14 elements? is the external subscription here just a composition of it's methods?

Library/ExternalSubscription.cs
```diff
@@ -0,0 +1,189 @@
...
+        public string Uuid { get; private set; }
+        public string AppIdentifier { get; private set; }
+        public bool? AutoRenew { get; set; }
+        public int Quantity { get; set; }
+        public ExternalProductReference ExternalProductReference { get; private set; }
+        public DateTime? LastPurchased { get; private set; }
+        public DateTime? ActivatedAt { get; private set; }
+        public DateTime? ExpiresAt { get; private set; }
+        public DateTime CreatedAt { get; private set; }
+        public DateTime UpdatedAt { get; private set; }
```
Patrick-Duvall: If top level nodes are `last_purchased, auto_renew, app_identifier, quantity, activated_at expires_at, created_at, updated_at` can we access Uuid Here?If yes, do you know where it's coming from?

Library/ExternalProduct.cs
ELepolt: That's a good point. For some reason I wasn't thinking and didn't think to follow the same way it's done in Subscription and give the user access to the whole plan.

Test/ExternalSubscriptionTest.cs
ELepolt: The `Quantity` is a field on ExternalSubscription. Not the quantity of external subscriptions received. 14 is the quantity I chose when I made this ExternalSubscription

Library/ExternalSubscription.cs
ELepolt: Meant to delete that.

Library/ExternalSubscription.cs
ELepolt: Are you asking for me to write a new class so that it's `externalSubscription.ExternalResource.ExternalObjectReference`?

Library/ExternalProduct.cs
ELepolt: Follows the same idea as Subscription that ultimately puts a PlanCode on the top level, but you can now call `subscription.Plan` and get everything

Library/ExternalSubscription.cs
ELepolt: You're right. That's the proper way to do it. That's been updated

Patrick-Duvall (approved):

PR #759, opened 2023-01-17
Add support for custom fields on charges

btruncali1:
Add support for custom fields on adjustments.  A few unit tests for purchases with custom fields.


douglaslise (approved):
LGTM!

Regarding to deprecations, are those warnings valid?

Issue #773, opened 2023-03-10
CancelSubscription does not pass in `body` for SubscriptionCancel for the request


MikeBairdRocks:
**Describe the bug**

`CancelSubscription` and `CancelSubscriptionAsync` do not pass in the body for `SubscriptionCancel` so the execution always defaults to `term_end`, when trying to use `bill_date` (https://recurly.com/developers/api/v2021-02-25/#operation/cancel_subscription) 

**To Reproduce**

Execute the `CancelSubscription` and `CancelSubscriptionAsync` with `Timeframe = Timeframe.BillDate`

```csharp
var subscriptionCancelRequest = new SubscriptionCancel
{
   Timeframe = Timeframe.BillDate
};
_recurlyClient.CancelSubscriptionAsync("mysubuuid", subscriptionCancelRequest, cancellationToken);
```

**Expected behavior**

The execution of CancelSubscription should result in the subscription ending on bill date, instead it is set to end of term.

**Your Environment**

- Which version of this library are you using?
`4.32.0`
- Which version of the language are you using?
`net7.0`

douglasmiller:
@MikeBairdRocks thank you for brining this to our attention (and finding the root cause + fix). 

The `Client.cs` file is automatically generated from our openapi specification, so I will not be able to merge your fix pull request. The good news is that I've updated our code generator to fix the bug and will get a new client generated soon. I'll let you know when it's published

douglasmiller:
@MikeBairdRocks The referenced pull request includes the fix for the CancelSubscription operation and a couple of others that also contain optional bodies.

I am publishing the [4.32.1](https://github.com/recurly/recurly-client-dotnet/releases/tag/4.32.1) version of the client library now.

Thanks again for letting us know about this issue!

PR #774, opened 2023-03-10
fix: Bug in not passing body to request for CancelSubscription

MikeBairdRocks:
Fix for ISSUE #773


douglasmiller:
Unable to merge since this is modifying generated code that would be overwritten. An update to the code generator will address this issue.

PR #802, opened 2023-07-17
[V2] Add starting_on and ending_on to subscription ramps

rlew421:
Adds `starting_on` and `ending_on` to `SubscriptionRampInterval`

### DIFF
--- a/Test/SubscriptionTest.cs
+++ b/Test/SubscriptionTest.cs
@@ -856,6 +856,7 @@ namespace Recurly.Test
             var subscription = new Subscription(account, plan, "USD");
             subscription.RampIntervals = GetMockSubscriptionRampIntervals(5);
             subscription.RampIntervals.Count.Should().Be(5);
+            subscription.StartsAt = 1.January(2099);
             subscription.Create();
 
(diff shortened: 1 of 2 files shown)


Test/SubscriptionTest.cs
```diff
@@ -861,8 +861,12 @@ public void CreateSubscriptionWithCustomRamps()
             var fromService = Subscriptions.Get(subscription.Uuid);
             fromService.RampIntervals.Count.Should().Be(5);
 
+            var now = DateTime.Now;
+
             fromService.RampIntervals[0].StartingBillingCycle.Should().Be(1);
             fromService.RampIntervals[0].UnitAmountInCents.Should().Be(500);
+            fromService.RampIntervals[0].StartingOn.Should().BeCloseTo(now, precision: 2000);
```
rlew421: I used this assertion instead of passing in a starting on date – it was a bit too involved to try to pass in a starting date in the example in `BaseTest` because it would need to create `StartingOn` and `EndingOn` with different months. Fluent assertions uses the string of the month name instead of an integer for the month in order to pass in a date. Mapping through that in this example seemed more involved than what the test calls for

csampson (approved):

PR #811, opened 2023-12-12
feat add event handlers to make request async

aron-ridgway-deltatre:
### Issue we are facing
The issue we are facing is we have a multi tenant application and we need to add tracing/monitoring for Recurly http calls per tenant, so we can monitor the amount off calls being made by each tenant which include paged requests that happen in the background.
BaseClient MakeRequest<T> has Event handlers which can be used to hook into http request and responses.
MakeRequestAsync<T> that we use in our service is currently missing the Event Handling processing,

### Proposed solution
Add the event handling logic  to the MakeRequestAsync<T> method.

There may be a better way to solve this issue, so we are open to your suggestions on the best implementation to resolve this issue.

### DIFF
--- a/Recurly.Tests/BaseClientTest.cs
+++ b/Recurly.Tests/BaseClientTest.cs
@@ -183,6 +183,23 @@ namespace Recurly.Tests
             Assert.Equal("benjamin", resource.MyString);
         }
 
+        [Fact]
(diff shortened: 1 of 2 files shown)


douglasmiller (approved):
Thank you for creating the pull request @aron-ridgway-deltatre!

PR #818, opened 2024-02-01
[V2] Add RevRec settings to the Plan entity

Smolations:
Adds the following properties to the `Plan` entity for the V2 client:
- `LiabilityGlAccountId`
- `RevenueGlAccountId`
- `PerformanceObligationId`
- `SetupFeeLiabilityGlAccountId`
- `SetupFeeRevenueGlAccountId`
- `SetupFeePerformanceObligationId`

### Examples

The following examples only show the base fee RevRec settings for brevity, but the `SetupFee*` properties behave the same way.
```csharp
## create a plan with default RevRec settings
var plan = new Plan();
# ...
plan.Create();
Console.WriteLine("Liability ID: {0}", plan.LiabilityGlAccountId); # => ""
Console.WriteLine("Revenue ID: {0}", plan.RevenueGlAccountId);     # => ""
Console.WriteLine("POB ID: {0}", plan.PerformanceObligationId);    # => "abc" (default POB)

## remove RevRec settings from a plan
var plan = Plans.Get(planCode);

# note that POB IDs cannot be removed, but they can be changed
plan.LiabilityGlAccountId = null;
plan.RevenueGlAccountId = null;
plan.Update();
```

### DIFF
--- a/Library/Plan.cs
+++ b/Library/Plan.cs
@@ -5,7 +5,7 @@ using System.Xml;
 
 namespace Recurly
 {
-    public class Plan : RecurlyEntity
+    public class Plan : RevRecEntity
     {
         public enum IntervalUnit
         {
@@ -40,6 +40,10 @@ namespace Recurly
 
         public string AccountingCode { get; set; }
         public string SetupFeeAccountingCode { get; set; }
+
+        public string SetupFeeLiabilityGlAccountId = "";
+        public string SetupFeeRevenueGlAccountId = "";
+        public string SetupFeePerformanceObligationId { get; set; }
 
         public DateTime CreatedAt { get; private set; }
         public DateTime UpdatedAt { get; private set; }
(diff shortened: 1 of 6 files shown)


Library/Plan.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Recurly
 {
-    public class Plan : RecurlyEntity
+    public class Plan : RevRecEntity
```
tilley-kyle, line 8: Why is this changing? Is there still a `RecurlyEntity` being used?

Smolations: Yes, if you look at the thing taking its place, `RevRecEntity`, you will see that _it_ is inheriting from `RecurlyEntity`, so the new class is just fitted "in between" the original two.

tilley-kyle: Ohhh okay. Cool. That makes sense.

tilley-kyle (approved):

8eth:
DevQA+1

PR #833, opened 2024-05-03
Add subscription change proflex settings

smagdicatrecurly:
Adds proration_settings to subscription attributes.

### DIFF
--- a/Library/ProrationSettings.cs
+++ b/Library/ProrationSettings.cs
@@ -1,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+
+namespace Recurly
+{
+    public class ProrationSettings
+    {
+        public enum Options
+        {
+            ProratedAmount,
+            FullAmount,
+            None
+        }
+
+        /// <summary>
+        /// Proration behavior to be applied to charges for the subscription change.
+        /// When not set, default proration settings will be applied.
+        /// </summary>
+        public Options? Charge { get; set; }
+
+        /// <summary>
+        /// Proration behavior to be applied to credits for the subscription change.
+        /// When not set, default proration settings will be applied.
+        /// </summary>
+        public Options? Credit { get; set; }
+    }
+}
--- a/Library/SubscriptionChange.cs
+++ b/Library/SubscriptionChange.cs
@@ -25,6 +25,12 @@ namespace Recurly
         public string PlanCode { get; set; }
 
         /// <summary>
+        /// Proration settings to be applied to the subscription change.
+        /// If not set, the default proration behavior will apply.
+        /// </summary>
+        public ProrationSettings ProrationSettings { get; set; }
+
+        /// <summary>
         /// List of custom fields
         /// </summary>
         public List<CustomField> CustomFields
@@ -126,6 +132,20 @@ namespace Recurly
             xmlWriter.WriteStringIfValid("billing_info_uuid", BillingInfoUuid);
 
             xmlWriter.WriteStringIfValid("plan_code", PlanCode);
+
+            if (ProrationSettings != null)
+            {
+                xmlWriter.WriteStartElement("proration_settings");
+
+                if (ProrationSettings.Charge != null)
+                    xmlWriter.WriteElementString("charge", ProrationSettings.Charge.Value.ToString().EnumNameToTransportCase());
+
+                if (ProrationSettings.Credit != null)
+                    xmlWriter.WriteElementString("credit", ProrationSettings.Credit.Value.ToString().EnumNameToTransportCase());
+
+                xmlWriter.WriteEndElement();
+            }
+
 
             if (AddOns != null)
                 xmlWriter.WriteIfCollectionHasAny("subscription_add_ons", AddOns);
--- a/Test/SubscriptionTest.cs
+++ b/Test/SubscriptionTest.cs
@@ -306,6 +306,54 @@ namespace Recurly.Test
         }
 
         [RecurlyFact(TestEnvironment.Type.Integration)]
+        public void UpdateSubscriptionWithProrationSettings()
+        {
+            var plan = new Plan(GetMockPlanCode(), GetMockPlanName())
+            {
+                Description = "Update Subscription Plan 1"
+            };
+            plan.UnitAmountInCents.Add("USD", 1500);
+            plan.Create();
+            PlansToDeactivateOnDispose.Add(plan);
+
+            var plan2 = new Plan(GetMockPlanCode(), GetMockPlanName())
+            {
+                Description = "Update Subscription Plan 2"
+            };
+            plan2.UnitAmountInCents.Add("USD", 750);
+            plan2.Create();
+            PlansToDeactivateOnDispose.Add(plan2);
+
+            var account = CreateNewAccountWithBillingInfo();
+
+            var sub = new Subscription(account, plan, "USD");
+            sub.Create();
+
+            var subChange = new SubscriptionChange()
+            {
+                PlanCode = plan2.PlanCode,
+                ProrationSettings = new ProrationSettings()
+                {
+                    Charge = ProrationSettings.Options.None,
+                    Credit = ProrationSettings.Options.FullAmount
+                }
+            };
+
+            Assert.Equal(subChange.ProrationSettings.Charge, ProrationSettings.Options.None);
+            Assert.Equal(subChange.ProrationSettings.Credit, ProrationSettings.Options.FullAmount);
+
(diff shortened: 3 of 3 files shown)


Library/SubscriptionChange.cs
```diff
@@ -24,6 +24,12 @@ public enum ChangeTimeframe : short
 
         public string PlanCode { get; set; }
 
+        /// <summary>
+        /// Proration settings to be applied to the subscription change.
+        /// If not set, the default proration behavior will apply.
+        /// </summary>
+        public ProrationSettings ProrationSettings { get; set; }
```
btruncali1, line 31: should this be nullable?

Test/SubscriptionTest.cs
```diff
@@ -305,6 +305,51 @@ public void UpdateSubscription()
...
+            var account = CreateNewAccountWithBillingInfo();
+
+            var sub = new Subscription(account, plan, "USD");
+            sub.Create();
+
+            var subChange = new SubscriptionChange()
+            {
+                PlanCode = plan2.PlanCode,
+                BillingInfoUuid = account.GetBillingInfos()[0].Id,
+                NetTerms = 10,
```
btruncali1: **nit:** you can just not set NetTerms at all here, or make it 0 since you're not testing it

Test/SubscriptionTest.cs
```diff
@@ -305,6 +305,51 @@ public void UpdateSubscription()
...
+                NetTermsType = NetTermsType.NET,
+                ProrationSettings = new ProrationSettings()
+                {
+                    Charge = ProrationSettings.Options.FullAmount,
+                    Credit = ProrationSettings.Options.None
+                }
+            };
+
+            Assert.Equal(subChange.ProrationSettings.Charge, ProrationSettings.Options.FullAmount);
```
btruncali1: To actually test this you probably want to do the assertions last, after you get the subscription here:
`var newSubscription = Subscriptions.Get(sub.Uuid);`

Then assert with the newSubscription:

` Assert.Equal(newSubscription.ProrationSettings.Charge, ProrationSettings.Options.FullAmount);`

Library/ProrationSettings.cs
```diff
@@ -0,0 +1,29 @@
...
+        public Options? Charge { get; set; }
+
+        /// <summary>
+        /// Proration behavior to be applied to credits for the subscription change.
+        /// Can be set to FullAmount, ProratedAmount, or None.
+        /// When not set, default proration settings will be applied.
+        /// </summary>
+        public Options? Credit { get; set; }
+    }
+}
```
btruncali1: EOF line

Library/SubscriptionChange.cs
smagdicatrecurly: Yes. The user can fallback to default behavior set at the site level or recurly's default behavior which is simply to prorate.

btruncali1:
Just confirming this is for SubscriptionChange only right?  They aren't able to create a new subscription with these settings?

smagdicatrecurly:
Thats correct. This feature allows a merchant to decide how they want to charge or credit a customer when a subscription is changed after the start of a billing cycle.

Library/SubscriptionChange.cs
btruncali1: Then you want to do something like this

`public ProrationSettings? ProrationSettings { get; set; }`

Test/SubscriptionTest.cs
smagdicatrecurly: This is not returned on the subscription. I am only able to check these on the subscription change. We pass it in but it is not passed back in the response.

Library/SubscriptionChange.cs
smagdicatrecurly: On second thought, this is just an empty object if a user instantiates this class. So no it shouldbe be nullable. What is nullable is Charge and Credit properties.

Test/SubscriptionTest.cs
```diff
@@ -305,6 +305,54 @@ public void UpdateSubscription()
...
+            Assert.Equal(subChange.ProrationSettings.Credit, ProrationSettings.Options.FullAmount);
+
+            Subscription.ChangeSubscription(sub.Uuid, subChange);
+
+            var newSubscription = Subscriptions.Get(sub.Uuid);
+            var invoices = account.GetInvoices();
+            var chargeInvoice = invoices.FirstOrDefault(invoice => invoice.Type == "charge");
+            var creditInvoice = invoices.FirstOrDefault(invoice => invoice.Type == "credit");
+
+            Assert.Equal(0, chargeInvoice.SubtotalInCents);
```
smagdicatrecurly, line 352: Added assertions to check proration settings applied to charges and credits here.

Library/SubscriptionChange.cs
smagdicatrecurly: I thought about this but with our version of C# it throws errors both when building the branch and testing building the client locally with the following error: `Feature 'nullable reference types' is not available in C# 7.3.`

I was able to update Recurly.csproj to the latest language version to allow nullable reference types and build successfully but I don't want to change our current configuration.

btruncali1: hmm thats really strange there are lots of nullable reference types all over the codebase, you shouldnt need to update the language version

btruncali1: Although, looking at this you probably just dont need to make this nullable because everything inside the model itself is nullable, so this is fine

Library/ProrationSettings.cs
```diff
@@ -0,0 +1,29 @@
...
+        public enum Options
+        {
+            ProratedAmount,
+            FullAmount,
+            None
+        }
+
+        /// <summary>
+        /// Proration behavior to be applied to charges for the subscription change.
+        /// Can be set to FullAmount, ProratedAmount, or None.
```
csampson: Technically, you don't need to document "can be set to...".

Library/SubscriptionChange.cs
csampson: @btruncali1 @smagdicatrecurly My understanding is that, because this is a reference type, it's nullable by default - compare dto value types (e.g. `int`), where you would have to declare nullability.

This thinking seems consistent with both the aforementioned compile error and being able to do `ProrationSettings != null` below.

Library/SubscriptionChange.cs
```diff
@@ -127,6 +133,20 @@ internal void WriteChangeSubscriptionXml(XmlTextWriter xmlWriter)
 
             xmlWriter.WriteStringIfValid("plan_code", PlanCode);
 
+            if (ProrationSettings != null)
+            {
+                xmlWriter.WriteStartElement("proration_settings");
+
+                if (ProrationSettings.Charge != null)
```
csampson, line 140: Ah, thanks for adding null safety here.

Test/SubscriptionTest.cs
```diff
@@ -305,6 +305,54 @@ public void UpdateSubscription()
...
+        {
+            var plan = new Plan(GetMockPlanCode(), GetMockPlanName())
+            {
+                Description = "Update Subscription Plan 1"
+            };
+            plan.UnitAmountInCents.Add("USD", 1500);
+            plan.Create();
+            PlansToDeactivateOnDispose.Add(plan);
+
+            var plan2 = new Plan(GetMockPlanCode(), GetMockPlanName())
```
csampson, line 319: Not critical, but you could make this test leaner by _just_ changing `UnitAmountInCents` on the change.

smagdicatrecurly: I was hoping to see BOTH credits and charges by changing the plan.

Library/ProrationSettings.cs
smagdicatrecurly: I thought it was helpful when hovering over that property on SubscriptionChange but if its not a common pattern to do this I can remove that.

smagdicatrecurly: I see it as a pattern actually to note the possible values.

Screen Shot 2024-05-07 at 12 56 29
Screen Shot 2024-05-07 at 12 56 52

csampson: I think that's fine for value types, such as `NetTerms`, but we and the consumer have access to the literal enum values.

Test/SubscriptionTest.cs
csampson: Ah, fair.

Library/ProrationSettings.cs
smagdicatrecurly: Ah got it. Right.

csampson (approved):

PR #858, opened 2025-01-22
 Fixing the ListEntitlements return type

### LINKED ISSUE #853: ListEntitlements method fails to deserialize the response data
**Describe the bug**

When calling the `ListEntitlements` method, the client returns a `Pager<Entitlements>` instead of a `Pager<Entitlement>` object. This causes improper deserialization of the response and fails to return the Entitlement data for an Account.

**To Reproduce**

Create an Entitlement, and have it assigned to an Account.

```
var entitlementPager = client.ListEntitlements("code-...");
var page = entitlementPager.FetchNextPage();
// See that `page.Data` is a `List<Entitlement>`, and that list does not contain the expected Entitlement
```

Taking the `RawResponse` of the `page`, we can manually deserialize it into a `Pager<Entitlement>` and successfully retrieve the data.

```
var raw = page.GetResponse().RawResponse;

var restResponse = new RestResponse
{
    Content = raw
};

var deserialized = JsonSerializer.Default.Deserialize<Pager<Recurly.Resources.Entitlement>>(restResponse);
```

**Expected behavior**

The client correctly deserializes the response from the API.

**Your Environment**

- Which version of this library are you using? 4.58.0

jweber:
I know the `IClient` and `Client` files are auto-generated, but this should show what's wrong with the return type for this method.

Fixing this should resolve #853.

### DIFF
--- a/Recurly/Client.cs
+++ b/Recurly/Client.cs
@@ -5190,7 +5190,7 @@ namespace Recurly
         /// <returns>
         /// A list of the entitlements granted to an account.
         /// </returns>
-        public Pager<Entitlements> ListEntitlements(string accountId, ListEntitlementsParams optionalParams = null, RequestOptions options = null)
+        public Pager<Entitlement> ListEntitlements(string accountId, ListEntitlementsParams optionalParams = null, RequestOptions options = null)
         {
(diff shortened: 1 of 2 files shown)


cbarton:
fixed in #908

PR #876, opened 2025-07-02
Add constant apple_pay_merchant_token

jsanderson1130:
When updating work on supporting payment methods, we need to update to utilize `apple_pay_merchant_token` moving forward.


cvelezprecurly (approved):
LGTM

PR #896, opened 2026-04-10
Remove RestSharp dependency

douglasmiller:
This change will remove the RestSharp dependency and instead leverage native .NET capabilities to submit requests and handle responses from the API.


Recurly.Tests/BaseClientTest.cs
```diff
@@ -130,13 +128,20 @@ public void WillIncludeCustomHeaders()
...
-            Func<IRestRequest, bool> matcher = delegate (IRestRequest request)
+            bool matcherCalled = false;
+            bool pathCorrect = false;
+
+            Func<HttpRequestMessage, bool> matcher = delegate (HttpRequestMessage request)
             {
-                Assert.Equal("/my_resources/douglas%2F", request.Resource);
+                matcherCalled = true;
+                pathCorrect = request.RequestUri.AbsolutePath.Contains("douglas%2F") ||
+                              request.RequestUri.AbsolutePath.Contains("douglas%252F");
```
btruncali1, line 138: Question:  how do we get this bottom one `"douglas%252F"`.  Is it just double encoded?  Seems like that request would fail if it hit the server, so im curious in what case we would get this, and why its allowed.

btruncali1 (dismissed):

btruncali1 (approved):

